PR #32, opened 2024-12-05
Add tests for CachedLoader

KarimTayie:
Adds tests to verify template caching, handling of missing templates, and invalid encoding errors, ensuring reliable CachedLoader functionality.

### DIFF
--- a/src/loaders.rs
+++ b/src/loaders.rs
@@ -266,7 +266,7 @@ mod tests {
             let template = cached_loader
                 .get_template(py, "basic.txt")
                 .expect("Failed to load template")
-                .expect("Template is None");
+                .expect("Template file could not be read");
 
             // Verify the template filename
             let mut expected_path = std::env::current_dir().expect("Failed to get current directory");
@@ -281,7 +281,7 @@ mod tests {
             let template = cached_loader
                 .get_template(py, "basic.txt")
                 .expect("Failed to load template")
-                .expect("Template is None");
+                .expect("Template file could not be read");
 
             // Verify the template filename again
             assert_eq!(template.filename.unwrap(), expected_path);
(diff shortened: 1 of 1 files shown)


LilyFoote:
Thanks for this! I'll give this a proper review this evening and look into the coverage upload issue.

src/loaders.rs
```diff
@@ -239,6 +239,104 @@ mod tests {
...
+                FileSystemLoader::new(vec!["tests/templates".to_string()], encoding_rs::UTF_8);
+
+            // Wrap the FileSystemLoader in a CachedLoader
+            let mut cached_loader = CachedLoader::new(vec![Loader::FileSystem(filesystem_loader)]);
+
+            // Load a template via the CachedLoader
+            let template = cached_loader
+                .get_template(py, "basic.txt")
+                .expect("Failed to load template")
+                .expect("Template is None");
```
LilyFoote, line 269: Currently the only way this second `Result` can be an error is if the template file has the wrong encoding, so I think this is more accurate:

```suggestion
                .expect("Template file could not be read");
```

src/loaders.rs
```diff
@@ -239,6 +239,104 @@ mod tests {
...
+
+            // Verify the cache state after first load
+            assert_eq!(cached_loader.cache.len(), 1);
+            verify_cache(&cached_loader.cache, "basic.txt", &expected_path);
+
+            // Load the same template again via the CachedLoader
+            let template = cached_loader
+                .get_template(py, "basic.txt")
+                .expect("Failed to load template")
+                .expect("Template is None");
```
LilyFoote, line 284:
```suggestion
                .expect("Template file could not be read");
```

LilyFoote (approved):
Thanks! This is great.

PR #34, opened 2024-12-05
Cache the LoaderError in CachedLoader

LilyFoote:
Django's cached loader caches the `TemplateDoesNotExist` exception. This corresponds to our `LoaderError`, so we should cache it too. This avoids repeating the template lookup when we already know we can't find it.

### DIFF
--- a/src/loaders.rs
+++ b/src/loaders.rs
@@ -123,7 +123,10 @@ impl CachedLoader {
                         Err(mut e) => tried.append(&mut e.tried),
                     }
                 }
-                Err(LoaderError { tried })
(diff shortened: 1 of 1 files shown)


LilyFoote:
@KarimTayie Thanks again for writing those tests in #32. That was really helpful for me spotting this missing bit of caching!

KarimTayie:
> @KarimTayie Thanks again for writing those tests in #32. That was really helpful for me spotting this missing bit of caching!

You're very welcome. I'm glad the tests helped spot the caching issue. It was a pleasure contributing to the project!

PR #50, opened 2025-02-15
Support html autoescaping

LilyFoote:


src/parse.rs
```diff
@@ -696,6 +862,19 @@ impl<'t, 'l, 'py> Parser<'t, 'l, 'py> {
...
+        at: (usize, usize),
+        parts: TagParts,
+    ) -> Result<TokenTree, PyParseError> {
+        let token = lex_autoescape_argument(self.template, parts).map_err(ParseError::from)?;
+        let nodes = self.parse_until(EndTagType::Autoescape, "autoescape", at)?;
+        Ok(TokenTree::Tag(Tag::Autoescape {
+            enabled: token.enabled,
+            nodes,
+        }))
+    }
```
jonathan-s, line 877: I think it would be helpful to decompose the parser class, such that the parsing belongs to each built-in tag. Though that would be a separate effort, and perhaps you're already thinking along those lines.

src/render.rs
```diff
@@ -135,14 +186,33 @@ impl<'t, 'py> IntoOwnedContent<'t, 'py> for String {
     }
 }
 
-pub trait Render {
+/// Trait for resolving a template element into content suitable for
+/// further processing by another template element.
```
jonathan-s, line 190: Great to have a clear distinction between resolve and render.

src/render.rs
```diff
@@ -135,14 +186,33 @@ impl<'t, 'py> IntoOwnedContent<'t, 'py> for String {
...
+pub trait Resolve {
     fn resolve<'t, 'py>(
         &self,
         py: Python<'py>,
         template: TemplateString<'t>,
         context: &mut Context,
     ) -> TemplateResult<'t, 'py>;
```
jonathan-s, line 197: I'm wondering if we shouldn't have the type (alias) be `ResolvedResult` here. Whereas for the `Render` trait and the `render` method result be `TemplateResult`, or perhaps `RenderedResult`.

jonathan-s (commented):
Some thoughts here. Not much more to add. It seems pretty clear to me. 👌🏼

src/parse.rs
```diff
@@ -532,12 +630,67 @@ impl<'t, 'l, 'py> Parser<'t, 'l, 'py> {
                         token.at.0 + START_TAG_LEN,
                     )?
                     .into(),
-                TokenType::Tag => self.parse_tag(token.content(self.template), token.at)?,
-            })
+                TokenType::Tag => match self.parse_tag(token.content(self.template), token.at)? {
+                    Either::Left(token_tree) => token_tree,
+                    Either::Right(end_tag) => {
```
jonathan-s, line 635: So parse_tag returns `Result<Either<TokenTree, EndTag>>`, `Either<a, b>` seems to work a bit like a tuple, in this instance. So my question, what is the advantage of this compared to matching with a tuple?

LilyFoote: I think a tuple here would be like using a struct, but `Either` is an enum.

src/parse.rs
LilyFoote: I've thought about this before, but last time I tried it wasn't straightforward, so I gave up.

PR #52, opened 2025-02-16
Adding addslashes as a new filter

jonathan-s:
Here is my attempt at adding the `addslashes` filter. Haven't worked with rust a lot. So please feel free to share anything that is worth keeping in mind while getting better at rust. 

I added a fixture `assert_render` that wraps the template rendering for a consistent experience going forward for that kind of tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ $ source .venv/bin/activate
 $ pip install -r requirements.txt
 ```
 
-To run the Python tests, build Django Rusty Templates in develop mode with maturin and then run pytest:
+To run the Python tests, build Django Rusty Templates in develop mode with maturin and then run pytest. Each change in rust needs a new execution of maturin develop.
 
 ```bash
 $ maturin develop
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -759,6 +759,8 @@ impl<'t, 'l, 'py> Parser<'t, 'l, 'py> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::template::django_rusty_templates::{EngineData, Template};
+    use pyo3::types::{PyDict, PyDictMethods};
 
     use crate::lex::common::LexerError;
 
@@ -1101,6 +1103,30 @@ mod tests {
             let mut parser = Parser::new(py, template.into(), &libraries);
             let error = parser.parse().unwrap_err().unwrap_parse_error();
             assert_eq!(error, ParseError::InvalidNumber { at: (11, 5).into() });
+        })
+    }
+
+    #[test]
+    fn test_filter_parse_addslashes() {
+        pyo3::prepare_freethreaded_python();
+
+        Python::with_gil(|py| {
+            let engine = EngineData::empty();
+            let template_string = "{{ foo|addslashes }}".to_string();
+            let context = PyDict::new(py);
+            context.set_item("bar", "").unwrap();
+            let template = Template::new_from_string(py, template_string, &engine).unwrap();
+            let result = template.render(py, Some(context), None).unwrap();
+
+            assert_eq!(result, "");
+
+            let context = PyDict::new(py);
+            context.set_item("foo", "").unwrap();
+            let template_string = "{{ foo|addslashes:invalid }}".to_string();
+            let error = Template::new_from_string(py, template_string, &engine).unwrap_err();
+
+            let error_string = format!("{error}");
+            assert!(error_string.contains("addslashes filter does not take an argument"));
         })
     }
 
@@ -1580,6 +1606,11 @@ mod tests {
             let cloned = add.clone_ref(py);
             assert_eq!(add, cloned);
             assert!(add.py_eq(&cloned, py));
+
+            let add_slashes = FilterType::AddSlashes;
+            let cloned = add_slashes.clone_ref(py);
+            assert_eq!(add_slashes, cloned);
+            assert!(add_slashes.py_eq(&cloned, py));
         })
     }
 }
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -158,6 +158,7 @@ impl PyEq for TagElement {
 #[derive(Debug)]
 pub enum FilterType {
     Add(Argument),
+    AddSlashes,
     Default(Argument),
     External(Py<PyAny>, Option<Argument>),
     Lower,
@@ -169,6 +170,7 @@ impl PartialEq for FilterType {
             (Self::Add(a), Self::Add(b)) => a == b,
             (Self::Default(a), Self::Default(b)) => a == b,
             (Self::Lower, Self::Lower) => true,
+            (Self::AddSlashes, Self::AddSlashes) => true,
             (Self::External(_, _), Self::External(_, _)) => false, // Can't compare PyAny to PyAny
             _ => false,
         }
@@ -179,6 +181,7 @@ impl CloneRef for FilterType {
     fn clone_ref(&self, py: Python<'_>) -> Self {
         match self {
             Self::Add(arg) => Self::Add(arg.clone()),
+            Self::AddSlashes => Self::AddSlashes,
             Self::Default(arg) => Self::Default(arg.clone()),
             Self::External(filter, arg) => Self::External(filter.clone_ref(py), arg.clone()),
             Self::Lower => Self::Lower,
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -62,7 +62,7 @@ $ source .venv/bin/activate
 $ pip install -r requirements.txt
 ```
 
-To run the Python tests, build Django Rusty Templates in develop mode with maturin and then run pytest:
+To run the Python tests, build Django Rusty Templates in develop mode with maturin and then run pytest. Each change in rust needs a new execution of maturin develop.
```
jonathan-s, line 65: In hindsight this makes sense, but for a person that isn't used to it it's an unnecessary hiccup.

tests/filters/test_addslashes.py
```diff
@@ -0,0 +1,63 @@
...
+        }
+    )
+    """
+    output = self.engine.render_to_string(
+        "addslashes01", {"a": "<a>'", "b": mark_safe("<a>'")}
+    )
+    self.assertEqual(output, r"<a>\' <a>\'")
+
+
+@pytest.mark.xfail(reason="autoescape not there yet")
```
jonathan-s, line 29: I added these as xfails for the time being, marking them as accepted failures, and thus a flag to go back to these tests later on.

jonathan-s:
I'll look over the coverage as well.

LilyFoote:
> I wonder if coverage for python makes sense?
> 
> We're not that interested in coverage of test files, so either exclude the test files or remove coverage for python.

I think coverage for test files has some value - if we copy/paste a test to edit, but forget to rename it, coverage will catch this. That said, I agree the coverage report isn't very helpful currently, so I've added https://github.com/LilyFoote/django-rusty-templates/pull/54 to hide 100% covered files from the output.

src/render.rs
```diff
@@ -229,6 +229,16 @@ impl Render for Filter {
                     }
                 }
             }
+            FilterType::AddSlashes => match left {
+                Some(content) => Some(Content::String(Cow::Owned(
+                    content
+                        .render(context)?
+                        .replace("\\", "\\\\")
```
LilyFoote, line 236: Like Python, [Rust also supports raw strings](https://doc.rust-lang.org/rust-by-example/std/str.html#literals-and-escapes):

```suggestion
                        .replace(r"\", r"\\")
```

src/render.rs
```diff
@@ -229,6 +229,16 @@ impl Render for Filter {
...
+                        .replace("\"", "\\\"")
```
LilyFoote, line 237:
```suggestion
                        .replace(r#"""#, r#"\""#)
```

Though I'm not entirely sure this one is easier to read.

src/render.rs
```diff
@@ -229,6 +229,16 @@ impl Render for Filter {
...
+                        .replace("'", "\\'"),
```
LilyFoote, line 238:
```suggestion
                        .replace("'", r"\'"),
```

tests/filters/test_addslashes.py
```diff
@@ -0,0 +1,63 @@
...
+
+    output = self.engine.render_to_string(
+        "addslashes02", {"a": "<a>'", "b": mark_safe("<a>'")}
+    )
+    self.assertEqual(output, r"&lt;a&gt;\&#x27; <a>\'")
+
+
+def test_quotes(assert_render):
+    template = "{{ a|addslashes }} and {{ b|addslashes }}"
+    expected = "\\\"double quotes\\\" and \\'single quotes\\'"
```
LilyFoote, line 45:
```suggestion
    expected = r"\"double quotes\" and \'single quotes\'"
```

tests/conftest.py
```diff
@@ -0,0 +1,21 @@
...
+@pytest.fixture
+def django_temp():
+    return engines["django"].from_string
+
+
+@pytest.fixture
+def assert_render(rusty, django_temp):
+    def temp(template, context, expected):
+        assert django_temp(template).render(context) == expected
+        assert rusty(template).render(context) == expected
```
LilyFoote, line 19: When `assert_render` fails, is it clear which of these asserts was the cause? I'm a bit cautious that we might try debugging Rust, when actually the Django assert has failed.

tests/filters/test_addslashes.py
```diff
@@ -0,0 +1,63 @@
...
+    context = {"a": mark_safe('"double quotes"'), "b": mark_safe("'single quotes'")}
+
+    assert_render(template, context, expected)
+
+
+def test_backslashes(assert_render):
+    template = "{{ a|addslashes }}"
+    context = {"a": r"\ : backslashes, too"}
+    expected = "\\\\ : backslashes, too"
```
LilyFoote, line 54:
```suggestion
    expected = r"\\ : backslashes, too"
```

src/parse.rs
```diff
@@ -179,6 +181,7 @@ impl CloneRef for FilterType {
     fn clone_ref(&self, py: Python<'_>) -> Self {
         match self {
             Self::Add(arg) => Self::Add(arg.clone()),
+            Self::AddSlashes => Self::AddSlashes,
```
LilyFoote, line 184: I would test this, and the `PartialEq` and `PyEq` traits, by adding to the `test_filter_type_clone_ref` test right at the end of this file:

```rust
            let add_slashes = FilterType::AddSlashes;
            let cloned = add_slashes.clone_ref(py);
            assert_eq!(add_slashes, cloned);
            assert!(add_slashes.py_eq(&cloned, py));
```

LilyFoote (commented):
This is looking great! Once the test coverage is sorted (no completely uncovered lines, partial line coverage is acceptable) and my other suggestions are resolved, I'll merge this.

tests/conftest.py
jonathan-s: Output when django fails. So I would say it's clear. 
```
    def temp(template, context, expected):
>       assert django_temp(template).render(context) == expected
E       assert '\\"double qu...le quotes\\\'' == '\\"double qu... q quotes\\\''
E
E         - \"double quotes\" and \'single q quotes\'
E         ?                                --
E         + \"double quotes\" and \'single quotes\'

tests/conftest.py:18: AssertionError
```

tests/conftest.py
```diff
@@ -0,0 +1,21 @@
...
+
+    return temp
```
LilyFoote, line 21: I found myself reading `temp` as "temporary" rather than "template":

```suggestion
    def render_template(template, context, expected):
        assert django_temp(template).render(context) == expected
        assert rusty(template).render(context) == expected

    return render_template
```

tests/conftest.py
LilyFoote: I'd probably go with `django_template`.

src/render.rs
jonathan-s: Fixed

src/render.rs
jonathan-s: I left this as is, as it didn't seem that much more readable to me.

src/render.rs
jonathan-s: fixed.

tests/filters/test_addslashes.py
jonathan-s: fixed.

jonathan-s:
@LilyFoote Alright! We're hitting all rust lines with during coverage testing. The coverage that's being complained over right now is the python test files. I rebased on main, so it seems like the python coverage is still being picked up.

jonathan-s:
> I think coverage for test files has some value - if we copy/paste a test to edit, but forget to rename it, coverage will catch this. That said, I agree the coverage report isn't very helpful currently, so I've added https://github.com/LilyFoote/django-rusty-templates/pull/54 to hide 100% covered files from the output.

Once this is merged I've got another PR in line that can address duplicate test function names specifically.

src/render.rs
```diff
@@ -552,6 +562,31 @@ user = User('Lily')
...
+            let template = TemplateString("{{ quotes|addslashes }}");
+            let variable = Variable::new((3, 6));
+            let filter = Filter {
+                at: (10, 10),
+                left: TagElement::Variable(variable),
+                filter: FilterType::AddSlashes,
+            };
+
+            let rendered = filter.render(py, template, &mut context).unwrap();
+            assert_eq!(rendered, "\\'hello\\'");
```
LilyFoote, line 586:
```suggestion
            assert_eq!(rendered, r"\'hello\'");
```

jonathan-s:
@LilyFoote Thanks for reviewing and merging! 🎉

PR #55, opened 2025-02-16
Add cargo-llvm into devs-dependency for convenience + update readme.md

jonathan-s:
Adding `cargo-llvm-cov` to dev-dependencies and expanding the readme with instructions on how to check coverage locally.


README.md
```diff
@@ -80,3 +80,13 @@ If you get an `ImportError` from python, you may need to set the `PYTHONPATH` en
...
+
+When submitting a PR we check coverage. You can check coverage locally with
+
+```bash
+$ cargo llvm-cov --workspace
+# generate html
+$ cargo llvm-cov --html
+# open it
+$ open -a <your_browser> target/llvm-cov/html/index.html
+```
```
LilyFoote, line 92: I think it makes sense to explain that we measure coverage from both Rust and Python tests:

````suggestion
```

Note this only checks the coverage generated by the Rust tests. We also measure coverage generated by the Python tests in CI, which means missing coverage locally isn't always an issue.
````

File: README.md [hunk 2]
```diff
@@ -80,3 +80,13 @@ If you get an `ImportError` from python, you may need to set the `PYTHONPATH` en
 ```bash
 export PYTHONPATH=/path/to/venv/lib/python3.x/site-packages
 ```
+
+When submitting a PR we check coverage. You can check coverage locally with
+
+```bash
+$ cargo llvm-cov --workspace
```
LilyFoote, line 87: What does `--workspace` add?

LilyFoote (commented):
Good idea, thanks!

README.md
jonathan-s: Actually looks like you can use `cargo llvm-cov` only. But for the record, `--workspace` would test all packages in the workspace.

PR #56, opened 2025-02-16
Use ruff linting; F811 will prevent duplicate test function names

jonathan-s:
Here we use linting to check that we don't get any duplicate function names in tests. [Specifically F811](https://docs.astral.sh/ruff/rules/#pyflakes-f) covers this. 

We get some other linting for free, and have the possibility to expand on linting rules if we want. 

So in that sense, perhaps using coverage for python test_*.py becomes redundant?

### DIFF
--- a/python/django_rusty_templates/__init__.py
+++ b/python/django_rusty_templates/__init__.py
@@ -4,6 +4,8 @@ from django.template.backends.django import get_installed_libraries
 
 from .django_rusty_templates import Engine, Template
 
+__all__ = [Engine, Template]
+
 
 class RustyTemplates(BaseEngine):
     app_dirname = "templates"
--- a/python/django_rusty_templates/__init__.py
+++ b/python/django_rusty_templates/__init__.py
@@ -4,7 +4,7 @@ from django.template.backends.django import get_installed_libraries
 
 from .django_rusty_templates import Engine, Template
 
-__all__ = [Engine, Template]
+__all__ = ["RustyTemplates", "Template"]
 
 
 class RustyTemplates(BaseEngine):
--- a/tests/filters/test_addslashes.py
+++ b/tests/filters/test_addslashes.py
@@ -4,7 +4,6 @@ https://github.com/django/django/blob/main/tests/template_tests/filter_tests/tes
 """
 
 import pytest
-from django.template import engines
 from django.utils.safestring import mark_safe
 
 
@@ -31,8 +30,8 @@ def test_addslashes02(self):
     """@setup({"addslashes02": "{{ a|addslashes }} {{ b|addslashes }}"})"""
     template = "{{ a|addslashes }} {{ b|addslashes }}"
 
-    django_template = engines["django"].from_string(template)
-    rust_template = engines["rusty"].from_string(template)
+    django_template = engines["django"].from_string(template)  # noqa
+    rust_template = engines["rusty"].from_string(template)  # noqa
 
     output = self.engine.render_to_string(
         "addslashes02", {"a": "<a>'", "b": mark_safe("<a>'")}
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -37,6 +37,9 @@ jobs:
       run: |
         source <(cargo llvm-cov show-env --export-prefix)
         .venv/bin/maturin develop
+    - name: Lint test files
+      run: |
+        .venv/bin/ruff check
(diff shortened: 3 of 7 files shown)


tests/filters/test_addslashes.py
```diff
@@ -29,10 +28,10 @@ def test_addslashes01(self):
 @pytest.mark.xfail(reason="autoescape not there yet")
 def test_addslashes02(self):
     """@setup({"addslashes02": "{{ a|addslashes }} {{ b|addslashes }}"})"""
-    template = "{{ a|addslashes }} {{ b|addslashes }}"
+    # template = "{{ a|addslashes }} {{ b|addslashes }}"
 
-    django_template = engines["django"].from_string(template)
-    rust_template = engines["rusty"].from_string(template)
+    # django_template = engines["django"].from_string(template)
+    # rust_template = engines["rusty"].from_string(template)
```
LilyFoote, line 34: Commenting this out defeats the point of `xfail` in my opinion.

python/django_rusty_templates/__init__.py
```diff
@@ -2,7 +2,8 @@
 from django.template.backends.base import BaseEngine
 from django.template.backends.django import get_installed_libraries
 
-from .django_rusty_templates import Engine, Template
+from .django_rusty_templates import Engine
+from .django_rusty_templates import Template as Template
```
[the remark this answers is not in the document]
jonathan-s: This is what's happening. 
```
python/django_rusty_templates/__init__.py:5:45: F401 `.django_rusty_templates.Template` imported but unused; consider removing, adding to `__all__`, or using a redundant alias
  |
3 | from django.template.backends.django import get_installed_libraries
4 |
5 | from .django_rusty_templates import Engine, Template
  |                                             ^^^^^^^^ F401
  |
  = help: Use an explicit re-export: `Template as Template`
  ```

File: tests/filters/test_addslashes.py (hunk 1 above)
Comment (jonathan-s) [reply tests/filters/test_addslashes.py]: I changed to using `# noqa` instead. Either way the test is failing.

**Comment by jonathan-s** [comment]
> I'd still prefer to keep measuring coverage for tests though

Fair enough, though there will likely be some false positives (ie coverage saying that we get reduced coverage) for some time if we go with using xfail. Which I think is a useful placeholder for tests that need fixing once they are possible.

File: python/django_rusty_templates/__init__.py (hunk 2 above)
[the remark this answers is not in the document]
Comment (LilyFoote) [reply python/django_rusty_templates/__init__.py:6]: Ah! Of the recommended solutions, I think the one we should actually use is to add `__all__`.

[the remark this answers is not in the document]
Comment (jonathan-s) [reply python/django_rusty_templates/__init__.py]: I changed to using `__all__` instead

File: python/django_rusty_templates/__init__.py [hunk 3]
```diff
@@ -4,6 +4,8 @@
 
 from .django_rusty_templates import Engine, Template
 
+__all__ = [Engine, Template]
```
LilyFoote, line 7: These should all be strings. I also think we want to export `RustyTemplates`, but not `Engine`.

```suggestion
__all__ = ["RustyTemplates", "Template"]
```

tests/filters/test_addslashes.py
```diff
@@ -31,8 +30,8 @@ def test_addslashes02(self):
     """@setup({"addslashes02": "{{ a|addslashes }} {{ b|addslashes }}"})"""
     template = "{{ a|addslashes }} {{ b|addslashes }}"
 
-    django_template = engines["django"].from_string(template)
-    rust_template = engines["rusty"].from_string(template)
+    django_template = engines["django"].from_string(template)  # noqa
+    rust_template = engines["rusty"].from_string(template)  # noqa
```
LilyFoote, line 34: I missed it when reviewing before, but I think the reason `ruff` is complaining is because `django_template` and `rust_template` are unused.

Both `test_addslashes01` and `test_addslashes02` refer to a `self.engine`, but this doesn't exist because these aren't methods on a unittest test class.

jonathan-s: See https://github.com/LilyFoote/django-rusty-templates/pull/57

python/django_rusty_templates/__init__.py
jonathan-s: Damn, you're right. Goes to show that I very rarely use `__all__`.

PR #57, opened 2025-02-16
Draft: Remove one xfail for addslashes that actually didn't use autoescape

jonathan-s:
Removing one xfail which didn't use autoescape. (though that test still fails, leaving this as a draft while I ponder on why).


LilyFoote:
> (though that test still fails, leaving this as a draft while I ponder on why).

I think this might be an ordering thing - is the variable being html escaped before having slashes added, where it should be slashes first and then html escaping?

jonathan-s:
> > (though that test still fails, leaving this as a draft while I ponder on why).
> 
> I think this might be an ordering thing - is the variable being html escaped before having slashes added, where it should be slashes first and then html escaping?

You are correct. 

```
            Some(content) => content
                .render(context)? <- culprit is here. 
                .replace(r"\", r"\\")
                .replace("\"", "\\\"")
```

PR #58, opened 2025-02-17
Implementing capfirst filter

jonathan-s:
Publishing draft.

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -268,13 +268,12 @@ impl Render for Filter {
                 Some(content) => {
                     let content_string = content.render(context)?.into_owned();
                     let mut chars = content_string.chars();
-                    let first_char = chars.next().unwrap_or('\0').to_uppercase();
+                    let first_char = match chars.next() {
+                        Some(c) => c.to_uppercase(),
+                        None => return Ok("".into_content()),
+                    };
                     let string: String = first_char.chain(chars).collect();
-                    if string == "\0" {
-                        "".into_content()
-                    } else {
-                        string.into_content()
-                    }
+                    string.into_content()
                 }
                 None => "".into_content(),
             },
--- a/src/render.rs
+++ b/src/render.rs
@@ -134,6 +134,32 @@ impl<'t, 'py> Content<'t, 'py> {
     }
 }
 
+trait IntoOwnedContent<'t, 'py> {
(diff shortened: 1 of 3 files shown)


src/render.rs
```diff
@@ -230,14 +262,17 @@ impl Render for Filter {
...
-                        .replace("\"", "\\\"")
-                        .replace("'", r"\'"),
-                ))),
-                None => Some(Content::String(Cow::Borrowed(""))),
+                Some(content) => content
+                    .render(context)?
+                    .replace(r"\", r"\\")
+                    .replace("\"", "\\\"")
+                    .replace("'", r"\'")
+                    .into_content(),
```
jonathan-s, line 270: The idea here with `into_content` was to make this easier to read, and also easier to turn something into the content that we expect.

tests/filters/test_capfirst.py
```diff
@@ -0,0 +1,40 @@
+"""
+Adapted from
+https://github.com/django/django/blob/5.1/tests/template_tests/filter_tests/test_capfirst.py
+"""
+
+import pytest
+from django.utils.safestring import mark_safe
+
+
+@pytest.mark.fail(reason="autoescape not ready yet")
```
LilyFoote, line 10: I'm not sure when I'll get to it, but I plan to add `{% autoescape %}` support to #50 before merging it.

jonathan-s:
@LilyFoote Now ready for review. We're hitting 100% coverage on the patch except for the python tests.

src/render.rs
jonathan-s: Now considering that perhaps we should name this `resolve` instead, and implement the render trait for those types.  Let me know what you think.

LilyFoote: I'd be interested to see how that looks, but perhaps that's best tried in a follow-up PR.

jonathan-s: Yeah, I'm not entirely certain if it's a good idea or not. Having naming consistency is worth something though.

jonathan-s:
@LilyFoote Ok, I think this ready to merge now.

LilyFoote (approved):
Amazing! Thanks again!

jonathan-s:
And thanks for reviewing! 🎉

PR #59, opened 2025-02-21
Filter refactor

jonathan-s:
This was certainly an interesting refactor. I was wondering if you perhaps have any advice when taking on refactors like these. For instance the first part where I introduced the unitstruct I just spent a lot of time changing all the places to accept unitstructs. Fairly easy, but tedious, are there strategies to make that less tedious? 

Or if you have any advice in general, which you find useful when working with refactors in rust.

### DIFF
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -1,82 +1,8 @@
-use crate::types::TemplateString;
-use num_bigint::BigInt;
+use crate::error::PyRenderError;
+use crate::render::{Context, IntoBorrowedContent, IntoOwnedContent, Render};
+use crate::types::Argument;
+use crate::{render::Content, types::TemplateString};
 use pyo3::prelude::*;
-
-struct PartsIterator<'t> {
-    variable: &'t str,
-    start: usize,
-}
-
-impl<'t> Iterator for PartsIterator<'t> {
-    type Item = (&'t str, (usize, usize));
-
-    fn next(&mut self) -> Option<Self::Item> {
-        if self.variable.is_empty() {
-            return None;
-        }
-
-        match self.variable.find('.') {
-            Some(index) => {
-                let part = &self.variable[..index];
-                let at = (self.start, index);
-                self.start += index + 1;
-                self.variable = &self.variable[index + 1..];
-                Some((part, at))
-            }
-            None => {
-                let part = self.variable;
-                self.variable = "";
-                Some((part, (self.start, part.len())))
-            }
-        }
-    }
-}
-
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub struct Text {
-    pub at: (usize, usize),
-}
-
-impl Text {
-    pub fn new(at: (usize, usize)) -> Self {
-        Self { at }
-    }
-}
-
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub struct Variable {
-    pub at: (usize, usize),
-}
-
-impl<'t> Variable {
-    pub fn new(at: (usize, usize)) -> Self {
-        Self { at }
-    }
-
-    pub fn parts(
-        &self,
-        template: TemplateString<'t>,
-    ) -> impl Iterator<Item = (&'t str, (usize, usize))> {
-        let start = self.at.0;
-        let variable = template.content(self.at);
-        PartsIterator { variable, start }
-    }
-}
-
-#[derive(Clone, Debug, PartialEq)]
-pub enum ArgumentType {
-    Variable(Variable),
-    Text(Text),
-    TranslatedText(Text),
-    Int(BigInt),
-    Float(f64),
-}
-
-#[derive(Clone, Debug, PartialEq)]
-pub struct Argument {
-    pub at: (usize, usize),
-    pub argument_type: ArgumentType,
-}
 
 #[derive(Debug)]
 pub enum FilterType {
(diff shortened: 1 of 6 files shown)


src/filters.rs
```diff
@@ -0,0 +1,272 @@
...
+    TranslatedText(Text),
+    Int(BigInt),
+    Float(f64),
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Argument {
+    pub at: (usize, usize),
+    pub argument_type: ArgumentType,
+}
```
jonathan-s, line 83: All of these types, it feels like they don't really belong here. Shall we move them to types.rs, or do you have other suggestions?

LilyFoote:
> This was certainly an interesting refactor. I was wondering if you perhaps have any advice when taking on refactors like these. For instance the first part where I introduced the unitstruct I just spent a lot of time changing all the places to accept unitstructs. Fairly easy, but tedious, are there strategies to make that less tedious?
> 
> Or if you have any advice in general, which you find useful when working with refactors in rust.

Nothing Rust specific. I think this is always going to be quite tedious. In a big enough codebase it might be worthwhile scripting something. In vim I might use a macro. In an IDE there's often refactoring tools which might work well. But none of these fully avoid the tedium in my experience.

src/parse.rs
```diff
@@ -1158,10 +1084,13 @@ mod tests {
...
-                    at: (15, 3),
-                    argument_type: ArgumentType::Variable(baz),
-                }),
+                filter: FilterType::Default(
+                    Argument {
+                        at: (15, 3),
+                        argument_type: ArgumentType::Variable(baz),
+                    },
+                    DefaultFilter,
+                ),
```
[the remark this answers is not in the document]
LilyFoote: You might have noticed that I'm actually writing a lot of tests in Python at this point. I'm not really certain what the right balance is though.

I'm definitely open to having more end to end or integration tests at the moment. Long term, if the CI or local testing runtime gets too high we can move back towards more unit tests.

Some of the more unit style tests (especially in `parse.rs`) were written early on to test general properties of the parser, rather than specific filters or tags.

src/filters.rs
```diff
@@ -0,0 +1,272 @@
...
+            None => None,
+        };
+        let filter = filter.bind(py);
+        let value = match arg {
+            Some(arg) => filter.call1((variable, arg))?,
+            None => filter.call1((variable,))?,
+        };
+        Ok(Some(Content::Py(value)))
+    }
+}
```
LilyFoote, line 164: I think we can probably simplify this down to one new trait:

```rust
pub trait ResolveFilter {
    fn resolve<'t, 'py>(
        &self,
        py: Python<'py>,
        template: TemplateString<'t>,
        context: &mut Context,
        left: Option<Content<'t, 'py>>,
    ) -> TemplateResult<'t, 'py>;
```

To do this, we'd need to move `Argument` into the `AddFilter` and `DefaultFilter` structs. And similarly move `Py<PyAny>` and `Option<Argument>` into `ExternalFilter`.

src/render.rs
```diff
@@ -31,35 +38,7 @@ pub enum Content<'t, 'py> {
...
-
-impl PyRenderError {
-    pub fn try_into_render_error(self) -> Result<RenderError, PyErr> {
-        match self {
-            Self::RenderError(err) => Ok(err),
-            Self::PyErr(err) => Err(err),
-        }
-    }
-}
+pub type TemplateResult<'t, 'py> = Result<Option<Content<'t, 'py>>, PyRenderError>;
```
LilyFoote, line 41: If you're comfortable with an interactive rebase, I think introducing `TemplateResult` would make a nice standalone commit before the rest of the refactor.

Not critical if it's too much hassle.

src/render.rs
```diff
@@ -232,72 +211,24 @@ impl Render for Filter {
...
-                        match left.add(right) {
-                            Ok(sum) => return Ok(Some(Content::Py(sum))),
-                            Err(_) => return Ok(None),
-                        }
-                    }
-                }
+        let result = match &self.filter {
+            FilterType::Add(right, AddFilter) => {
+                AddFilter::apply(left, right, py, template, context)
             }
```
LilyFoote, line 219: With `ResolveFilter` this would become:

```suggestion
            FilterType::Add(filter) => filter.resolve(py, template, context, left),
```

src/filters.rs
LilyFoote: I think if they're only used in one file they should stay there. But if they're used in multiple places it probably makes sense to define them in a standard location. `types.rs` seems reasonable.

src/render.rs
LilyFoote: If you do this, I think it would also be neater if code moves were in their own commits instead of mixed with refactoring. (I.e. refactor within a file as one commit, then move it to a new file as a second commit.)

Again this is a nice to have if not too big a pain.

LilyFoote (changes requested):
Thanks for working this through, I think it's an excellent idea!

I think with a few tweaks we can improve it even more:

src/filters.rs
jonathan-s: Moved these to types, as they were used in at least two files.

src/filters.rs
jonathan-s: I agree this is a lot nicer.

src/render.rs
jonathan-s: `TemplateResult` got it's own commit, and I agree that having the move could have been nice in its own commit. But seems like the payoff for the effort was a little to small this time, especially since it's nice to have each commit properly compile, which all commits now do.

src/parse.rs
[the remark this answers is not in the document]
jonathan-s: > Some of the more unit style tests (especially in parse.rs) were written early on to test general properties of the parser, rather than specific filters or tags.

Makes sense.

PR #61, opened 2025-02-23
Slugify filter

jonathan-s:
Adding the regex and unicode-normalization crate in this PR. Seems like it would be difficult to solve this without these crates, so including them looks like it is the right thing to do. 

The commits here are more atomic. I don't know if you have any preference regarding atomicity. Either way, happy to squash commits.

### DIFF
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -2,6 +2,8 @@ use crate::render::{Context, IntoBorrowedContent, IntoOwnedContent, Render, Temp
 use crate::types::Argument;
 use crate::{render::Content, types::TemplateString};
 use pyo3::prelude::*;
+use regex::Regex;
+use unicode_normalization::UnicodeNormalization;
 
 #[derive(Debug)]
 pub enum FilterType {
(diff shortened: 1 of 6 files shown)


src/filters.rs
```diff
@@ -197,3 +200,38 @@ impl ResolveFilter for LowerFilter {
...
+impl ResolveFilter for SlugifyFilter {
+    fn resolve<'t, 'py>(
+        &self,
+        variable: Option<Content<'t, 'py>>,
+        py: Python<'py>,
+        _template: TemplateString<'t>,
+        _context: &mut Context,
+    ) -> TemplateResult<'t, 'py> {
+        let non_word_re = Regex::new(r"[^\w\s-]").unwrap();
+        let whitespace_re = Regex::new(r"[-\s]+").unwrap();
```
jonathan-s, line 216: I know that you shouldn't be using unwrap in production code in general. Though, in this case this should be safe, as the input will never change. 

I also realized that we can't really use constants for regex at the top of the module which is a common pattern in python. I did see that there is a crate `lazy-static` that would help you with that, but perhaps it seems a bit overkill to include another dependency for that.

LilyFoote: For cases like this, I'd go with `.expect` to explain why it won't panic.

LilyFoote: As for lazy-static, I understand that's considered obsolete now. Instead we can use something like [`LazyLock`](https://doc.rust-lang.org/stable/std/sync/struct.LazyLock.html).

LilyFoote, line 216: Can we have some comments explaining how these regexes work?

src/filters.rs
```diff
@@ -197,3 +200,38 @@ impl ResolveFilter for LowerFilter {
...
+
+        let content = match variable {
+            Some(content) => {
+                let content = content.to_py(py).str()?.extract::<String>()?;
```
LilyFoote, line 220: This feels inefficient to me - we're converting to a Python object and then back to `String`. Can we do:

```suggestion
                let content = match content.resolve_string()?.content();
```

instead? Or is there a reason this is invalid?

LilyFoote (commented):
Thanks again!

I have a few questions and suggestions below.

I'd also like to land #50 before this.

PR #63, opened 2025-02-24
Split up render.rs into submodules

LilyFoote:
I think `render.rs` was getting a bit too large, so I've broken it up into some submodules.


jonathan-s (approved):
Looks good to me.

PR #66, opened 2025-02-27
Update build wheels workflow

LilyFoote:
Generated with `maturin generate-ci github`.


LilyFoote:
I discovered that we're slightly out of sync with the latest CI that `maturin generate-ci github` generates, so this updates that.

This doesn't fix the build failures on main - they will need to wait for the next GitHub image rollout (https://github.com/actions/runner-images/issues/11637).

Issue #60, opened 2025-02-23
Add error message to UrlLexerError::IncompleteKeywordArgument


LilyFoote:
Currently the error message is an empty string:
https://github.com/LilyFoote/django-rusty-templates/blob/ee5481f53577b8eaa3982f1d2fadfa07964720b5/src/lex/url.rs#L51

We can and should make this more useful for debugging.

siddheshzz:
Hi I tried to look into this, i think we can have something like this - 
```
 #[error("Incomplete keyword argument at {:?} to {} ", .at.offset(), .at.len()+.at.offset())]
    IncompleteKeywordArgument {
        #[label("here")]
        at: SourceSpan,
    },
```
Does this sounds good or we need something different/custom like `LexerError`

LilyFoote:
Hi @siddheshzz, thanks for taking a look at this. I think we probably only need "Incomplete keyword argument". The `miette` library takes care of showing where in the template the error is automatically.

I'd also suggest taking a look at this test:

https://github.com/LilyFoote/django-rusty-templates/blob/0f561f806ee275bf389d1f141a6b95e1f00e19c9/tests/tags/test_url.py#L145-L168

and seeing if you can copy and adapt it to this error.

siddheshzz:
Hi thanks for your time and reply, could you please elaborate on the last part. 
Do we need to have a error message something like `"Argument for '' not found. '' is not a complete keyword argument"`.

LilyFoote:
I just mean we need a test for the error message and gave an example of an existing test you can copy.

PR #67, opened 2025-02-27
Add error message to UrlLexerError::IncompleteKeywordArgument #60

siddheshzz:
Added `Incomplete keyword argument` error message


LilyFoote (changes requested):
This is a good start, thanks!

We need a test too though.

PR #68, opened 2025-02-28
Add appdirs loader implementation with tests

KarimTayie:
For #3 

This PR updates the appdirs loader code and adds four new tests. When running tests concurrently, they sometimes fail with:
`NameError("name 'sys' is not defined")`
This random issue might stem from a race condition during the django.setup() call.

### DIFF
--- a/src/loaders.rs
+++ b/src/loaders.rs
@@ -31,6 +31,17 @@ fn safe_join(directory: &Path, template_name: &str) -> Option<PathBuf> {
     }
 }
 
+fn get_app_template_dir(path: Bound<'_, PyAny>, dirname: &str) -> Result<Option<PathBuf>, PyErr> {
+    if path.is_truthy()? {
+        let path_buf: PathBuf = path.extract()?;
+        let template_path = path_buf.join(dirname);
+        if template_path.is_dir() {
+            return Ok(Some(template_path));
+        }
+    }
+    Ok(None)
+}
+
 #[cached(
     size = 128,          // Cache size
     result = true,       // Cache Result type
@@ -45,14 +56,10 @@ fn get_app_template_dirs(py: Python<'_>, dirname: &str) -> Result<Vec<PathBuf>, 
     let mut template_dirs = Vec::new();
     for app_config_result in app_configs.try_iter()? {
         let path = app_config_result?.getattr("path")?;
-        if path.is_truthy()? {
-            let path_str: String = path.extract()?;
-            let template_path = PathBuf::from(path_str).join(dirname);
-            if template_path.is_dir() {
+        if let Some(template_path) = get_app_template_dir(path, dirname)? {
                 template_dirs.push(template_path);
             }
         }
-    }
 
     Ok(template_dirs)
 }
@@ -248,6 +255,7 @@ impl Loader {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use pyo3::{BoundObject, IntoPyObjectExt};
 
     use quickcheck::quickcheck;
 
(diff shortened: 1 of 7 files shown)


src/loaders.rs
```diff
@@ -213,6 +267,23 @@ mod tests {
...
+                "setdefault",
+                ("DJANGO_SETTINGS_MODULE", "tests.settings"),
+                None,
+            )
+            .unwrap();
+
+        // Import the django module and call django.setup()
+        let django_module = PyModule::import(py, "django").unwrap();
+        django_module.call_method0("setup").unwrap();
+    }
```
LilyFoote, line 285: I think it would probably be simpler to use Python tests than try to configure Django from Rust. Is there a reason you preferred this approach?

src/loaders.rs
```diff
@@ -79,16 +119,30 @@ impl FileSystemLoader {
...
-        todo!()
+        let dirs_result = get_app_template_dirs(py, "templates");
+        let dirs = match dirs_result {
+            Ok(dirs) => dirs,
+            Err(e) => Vec::new(),
+        };
+
+        let filesystem_loader = FileSystemLoader::new(dirs, self.encoding);
+
+        filesystem_loader.get_template(py, template_name, engine)
```
LilyFoote, line 145: We can write this a bit more concisely with `unwrap_or_default()`:

```suggestion
        let dirs = get_app_template_dirs(py, "templates").unwrap_or_default();
        let filesystem_loader = FileSystemLoader::new(dirs, self.encoding);
        filesystem_loader.get_template(py, template_name, engine)
```

However, I wonder if we should be raising some or all of the errors from `get_app_template_dirs` up to the user:

```suggestion
        let dirs = get_app_template_dirs(py, "templates")?;
        let filesystem_loader = FileSystemLoader::new(dirs, self.encoding);
        filesystem_loader.get_template(py, template_name, engine)
```

src/loaders.rs
```diff
@@ -31,6 +32,45 @@ fn safe_join(directory: &Path, template_name: &str) -> Option<PathBuf> {
...
+    // Import django.apps
+    let apps_module = PyModule::import(py, "django.apps")?;
+
+    // Get the 'apps' object
+    let apps = apps_module.getattr("apps")?;
+
+    // Call get_app_configs() on apps
+    let app_configs = apps.call_method0("get_app_configs")?;
+
+    // Convert to iterator and process each app_config
```
LilyFoote, line 51: I don't think we need the comments here:

```suggestion
    let apps_module = PyModule::import(py, "django.apps")?;
    let apps = apps_module.getattr("apps")?;
    let app_configs = apps.call_method0("get_app_configs")?;
```

src/loaders.rs
```diff
@@ -31,6 +32,45 @@ fn safe_join(directory: &Path, template_name: &str) -> Option<PathBuf> {
...
+                    let path_str: String = path.extract()?;
+                    let template_path = PathBuf::from(path_str).join(dirname);
+                    if template_path.is_dir() {
+                        template_dirs.push(template_path.to_string_lossy().into_owned());
+                    }
+                }
+            }
+            Err(e) => return Err(e),
+        }
+    }
```
LilyFoote, line 69: We can inline some code here and use the `?` operator more:

```suggestion
    let mut template_dirs = Vec::new();
    for app_config_result in app_configs.try_iter()? {
        let path = app_config_result?.getattr("path")?;
        if path.is_some() {
            let path_str: String = path.extract()?;
            let template_path = PathBuf::from(path_str).join(dirname);
            if template_path.is_dir() {
                template_dirs.push(template_path.to_string_lossy().into_owned());
            }
        }
    }
```

src/loaders.rs
```diff
@@ -31,6 +32,45 @@ fn safe_join(directory: &Path, template_name: &str) -> Option<PathBuf> {
     }
 }
 
+#[cached(
+    size = 128,          // Cache size
+    result = true,       // Cache Result type
+    key = "String",      // Use owned String as key
+    convert = r##"{ dirname.to_string() }"## // Convert &str to String
+)]
```
LilyFoote, line 40: What is the goal of this caching? I wonder if this is premature optimisation.

src/loaders.rs
```diff
@@ -79,16 +119,30 @@ impl FileSystemLoader {
...
+        engine: &EngineData,
     ) -> Result<PyResult<Template>, LoaderError> {
-        todo!()
+        let dirs_result = get_app_template_dirs(py, "templates");
+        let dirs = match dirs_result {
+            Ok(dirs) => dirs,
+            Err(e) => Vec::new(),
+        };
+
+        let filesystem_loader = FileSystemLoader::new(dirs, self.encoding);
```
LilyFoote, line 143: I see that `FileSystemLoader::new` takes a `Vec<String>` and I assume that's why you made `get_app_template_dirs` return `Result<Vec<String>, PyErr>`. But `FileSystemLoader` stores `dirs` as a `Vec<PathBuf>`, which would be a more natural return type from `get_app_template_dirs`. For this reason, I think we should add a second constructor method to `FileSystemLoader` that takes a `Vec<PathBuf>` instead of `Vec<String>`.

LilyFoote (changes requested):
Thanks for making a start on this!

I have some suggestions and questions before I think this will be ready to merge:

src/loaders.rs
LilyFoote: Ah, I see that [Django includes an `lru_cache`](https://github.com/django/django/blob/ea1e3703bee28bfbe4f32ceb39ad31763353b143/django/template/utils.py#L97) in its implementation. This probably is worth caching then.

src/loaders.rs
KarimTayie: I used this approach to be able to test rust code following the old tests patterns. get_app_template_dirs calls get_app_configs from django which needs setup to be called from django first.
so in each test that use get_app_template_dirs I call setup_django to make sure that get_app_configs is called without any errors.

KarimTayie:
Thanks for all your comments! I've addressed them all, and I also managed to resolve the failing tests in the pipeline. Let me know if there's anything else you'd like me to adjust.

src/loaders.rs
```diff
@@ -78,16 +109,24 @@ impl FileSystemLoader {
...
         &self,
-        _py: Python<'_>,
-        _template_name: &str,
-        _engine: &EngineData,
+        py: Python<'_>,
+        template_name: &str,
+        engine: &EngineData,
     ) -> Result<PyResult<Template>, LoaderError> {
-        todo!()
+        let dirs = get_app_template_dirs(py, "templates").unwrap_or_default();
```
LilyFoote, line 127: I think it's more correct to raise here than to swallow errors:

```suggestion
        let dirs = match get_app_template_dirs(py, "templates") {
            Ok(dirs) => dirs,
            Err(e) => return Ok(Err(e))
        };
```

Note that we can't use `?` because we want to return a `PyErr`, not a `LoaderError`.

LilyFoote (changes requested):
Thanks for coming back to this, I think this is getting really close now!

I have two more things I think we still need to change:

LilyFoote:
> the tests sometimes fail with NameError("name 'sys' is not defined")

I haven't managed to replicate this one myself, so I'm not sure exactly what is going on. I think it's more likely to be a problem in our testing than a bug in the main code that we need to fix before landing this.

KarimTayie:
> > the tests sometimes fail with NameError("name 'sys' is not defined")
> 
> I haven't managed to replicate this one myself, so I'm not sure exactly what is going on. I think it's more likely to be a problem in our testing than a bug in the main code that we need to fix before landing this.

Yes, I also believe this issue is specific to the tests rather than a bug in the main code.

LilyFoote (approved):
Amazing! Thanks!

PR #70, opened 2025-03-01
Add the slugify filter 

jonathan-s:
Supersedes #61 as it was easier to do this in new commits rather than rebase the thing. 

Also an open question regarding file structure. When working through the files I just noticed that we have `render/filters.rs` and `filters.rs`, to me it would make sense to keep this in the same place. Or did you have any particular reasons for splitting that up?

### DIFF
--- a/src/render/filters.rs
+++ b/src/render/filters.rs
@@ -267,14 +267,14 @@ impl ResolveFilter for SlugifyFilter {
         variable: Option<Content<'t, 'py>>,
         py: Python<'py>,
         _template: TemplateString<'t>,
-        context: &mut Context,
+        _context: &mut Context,
     ) -> ResolveResult<'t, 'py> {
         let non_word_re = Regex::new(r"[^\w\s-]").unwrap();
         let whitespace_re = Regex::new(r"[-\s]+").unwrap();
 
         let content = match variable {
             Some(content) => {
-                let content = content.resolve_string(context)?.content();
+                let content = content.to_py(py)?.str()?.extract::<String>()?;
                 let content = content
                     .nfkd()
                     // first decomposing characters, then only keeping
--- a/src/render/filters.rs
+++ b/src/render/filters.rs
@@ -11,6 +11,8 @@ use crate::parse::Filter;
 use crate::render::types::{Content, Context};
 use crate::render::{Resolve, ResolveResult};
 use crate::types::TemplateString;
+use regex::Regex;
+use unicode_normalization::UnicodeNormalization;
 
 trait IntoOwnedContent<'t, 'py> {
     fn into_content(self) -> Option<Content<'t, 'py>>;
@@ -263,11 +265,31 @@ impl ResolveFilter for SlugifyFilter {
     fn resolve<'t, 'py>(
         &self,
         variable: Option<Content<'t, 'py>>,
-        _py: Python<'py>,
+        py: Python<'py>,
         _template: TemplateString<'t>,
         context: &mut Context,
     ) -> ResolveResult<'t, 'py> {
-        Ok("".as_content())
+        let non_word_re = Regex::new(r"[^\w\s-]").unwrap();
+        let whitespace_re = Regex::new(r"[-\s]+").unwrap();
+
+        let content = match variable {
+            Some(content) => {
+                let content = content.resolve_string(context)?.content();
+                let content = content
+                    .nfkd()
+                    // first decomposing characters, then only keeping
+                    // the ascii ones, filtering out diacritics for example.
+                    .filter(|c| c.is_ascii())
+                    .collect::<String>()
+                    .to_lowercase();
+                let content = non_word_re.replace_all(&content, "");
+                let content = content.trim();
+                let content = whitespace_re.replace_all(&content, "-");
+                content.to_string().into_content()
+            }
+            None => "".as_content(),
+        };
+        Ok(content)
     }
 }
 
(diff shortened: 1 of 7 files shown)


LilyFoote:
> Also an open question regarding file structure. When working through the files I just noticed that we have `render/filters.rs` and `filters.rs`, to me it would make sense to keep this in the same place. Or did you have any particular reasons for splitting that up?

I like having a split between code relevant to parsing and code relevant to rendering. Probably `filters.rs` should be `parse/filters.rs`, but I haven't dug into refactoring `parse.rs` yet.

jonathan-s:
Thanks for the context!

src/render/filters.rs
```diff
@@ -258,6 +270,35 @@ impl ResolveFilter for SafeFilter {
...
+                let content = NON_WORD_RE.replace_all(&content, "");
+                let content = content.trim();
+                let content = WHITESPACE_RE.replace_all(&content, "-");
+                content.to_string().into_content()
+            }
+            None => "".as_content(),
+        };
+        Ok(content)
+    }
+}
```
LilyFoote, line 300: Two suggestions in one here:

1. Pull out a `slugify` helper function.
2. Handle the different `content` variants explicitly, so we can avoid converting into Python unnecessarily.

I also discovered the existence of [`GILOnceCell::import`](https://docs.rs/pyo3/latest/pyo3/sync/struct.GILOnceCell.html#method.import), which can be used to avoid importing `SafeData` more than once.

We'll need tests to cover the extra `Content` match arms.

```suggestion
fn slugify(content: Cow<str>) -> Cow<str> {
    let content = content
        .nfkd()
        // first decomposing characters, then only keeping
        // the ascii ones, filtering out diacritics for example.
        .filter(|c| c.is_ascii())
        .collect::<String>()
        .to_lowercase();
    let content = NON_WORD_RE.replace_all(&content, "");
    let content = content.trim();
    let content = WHITESPACE_RE.replace_all(&content, "-");
    Cow::Owned(content.to_string())
}

static SAFEDATA: GILOnceCell<Py<PyType>> = GILOnceCell::new();

impl ResolveFilter for SlugifyFilter {
    fn resolve<'t, 'py>(
        &self,
        variable: Option<Content<'t, 'py>>,
        py: Python<'py>,
        _template: TemplateString<'t>,
        _context: &mut Context,
    ) -> ResolveResult<'t, 'py> {
        let content = match variable {
            Some(content) => match content {
                Content::Py(content) => {
                    let slug = slugify(Cow::Owned(content.str()?.extract::<String>()?));
                    #[allow(non_snake_case)]
                    let SafeData = SAFEDATA.import(py, "django.utils.safestring", "SafeData")?;
                    match content.is_instance(SafeData)? {
                        true => Some(Content::HtmlSafe(slug)),
                        false => Some(Content::String(slug)),
                    }
                }
                Content::Int(content) => {
                    Some(Content::String(slugify(Cow::Owned(content.to_string()))))
                }
                Content::Float(content) => {
                    Some(Content::String(slugify(Cow::Owned(content.to_string()))))
                }
                Content::String(content) => Some(Content::String(slugify(content))),
                Content::HtmlSafe(content) => Some(Content::HtmlSafe(slugify(content))),
            },
            None => "".as_content(),
        };
        Ok(content)
    }
}
```

LilyFoote (commented):
Thanks for persevering with this!

I think we can do better with avoiding unnecessary conversions to Python objects though:

src/render/filters.rs
jonathan-s: Thanks, I've applied this suggestion, and will make some refinements + add tests for the content match arms.

src/render/filters.rs
```diff
@@ -279,20 +297,22 @@ impl ResolveFilter for SlugifyFilter {
...
+                Content::Py(content) => {
+                    let slug = slugify(Cow::Owned(content.str()?.extract::<String>()?));
+                    #[allow(non_snake_case)]
+                    let SafeData = SAFEDATA.import(py, "django.utils.safestring", "SafeData")?;
+                    match content.is_instance(SafeData)? {
+                        true => Some(Content::HtmlSafe(slug)),
+                        false => Some(Content::String(slug)),
+                    }
+                }
+                // Int and Float requires no slugify, we only need to turn it into a string.
```
LilyFoote, line 310: Good point!

src/render/filters.rs
```diff
@@ -401,6 +401,15 @@ mod tests {
 
             assert_eq!(result, "a-b");
 
+            let engine = EngineData::empty();
+            let template_string = "{{ var|slugify }}".to_string();
+            let context = PyDict::new(py);
+            context.set_item("var", 1).unwrap();
```
jonathan-s, line 407: @LilyFoote It appears that even if we add an integer here this is still effectively treated as if this coming from the python arm. Do you have any thoughts around this?

LilyFoote: Yeah, `context` is basically a `HashMap<String, Py<PyAny>>`, so putting an integer in will convert to a Python `int`.

Instead, the way to do this is to chain multiple filters. I usually use `default` for this:

```
{{ var|default:1|slugify }}
```

The result of the `default` filter will be a `Content::Int(1)` if `var` is not defined.

jonathan-s: I see, that means that for the majority of cases this won't matter much. Though I guess the type of `HashMap` may change at some point to make this more relevant.

src/render/filters.rs
```diff
@@ -299,6 +360,72 @@ mod tests {
...
+            context.set_item("var", "a & b").unwrap();
+            let template = Template::new_from_string(py, template_string, &engine).unwrap();
+            let result = template.render(py, Some(context), None).unwrap();
+
+            assert_eq!(result, "a-b");
+
+            let engine = EngineData::empty();
+            let template_string = "{{ var|default:1|slugify }}".to_string();
+            let context = PyDict::new(py);
+            context.set_item("var", 1).unwrap();
```
LilyFoote, line 407:
```suggestion
```

jonathan-s: Noticed :)

src/render/filters.rs
```diff
@@ -299,6 +360,71 @@ mod tests {
         })
     }
 
+    #[test]
+    fn test_render_filter_slugify() {
```
LilyFoote, line 364: I think we should probably split this particular test up into multiple tests. If anything breaks for whatever reason, I'd rather see which of these scenarios fail rather than just one.

src/render/filters.rs
```diff
@@ -299,6 +360,71 @@ mod tests {
...
+            assert_eq!(result, "hello-world");
+
+            let engine = EngineData::empty();
+            let template_string = "{{ var|slugify }}".to_string();
+            let context = PyDict::new(py);
+            context.set_item("var", " hello world").unwrap();
+            let template = Template::new_from_string(py, template_string, &engine).unwrap();
+            let result = template.render(py, Some(context), None).unwrap();
+
+            assert_eq!(result, "hello-world");
```
LilyFoote, line 384: Splitting the tests up would also allow the test name to point to the difference between these two tests - the leading whitespace in the second.

jonathan-s: Agreed, collecting them in the same test was partially laziness, partially keeping them a bit more organized compared to have single test functions. 

All the tests have been broken up now though.

src/render/filters.rs
```diff
@@ -267,7 +328,23 @@ mod tests {
...
+            Some(mark_safe) => mark_safe,
+            None => {
+                let py_mark_safe = py.import("django.utils.safestring")?;
+                let py_mark_safe = py_mark_safe.getattr("mark_safe")?;
+                MARK_SAFE.set(py, py_mark_safe.into());
+                MARK_SAFE.get(py).unwrap()
+            }
+        };
+        let safe_string = mark_safe.call1(py, (string,))?;
+        Ok(safe_string)
```
jonathan-s, line 346: So generally, this would the pattern to call python functions from rust?

LilyFoote: I think in tests its a bit overkill to use `GILOnceCell` - that's more of an optimisation.

jonathan-s: Fair enough :). I'll leave it as is, as all things considering it works. 

Either way, this should be good to go now!

jonathan-s:
🎉

Issue #71, opened 2025-03-01
Context is a HashMap<String, Py<PyAny>>


jonathan-s:
Since the variables are `Py<PyAny>>` they are treated as python variables at first, and only if they are further processed they get rust types.

It would be ideal to have the context considered to be each correct type directly so that when variables are resolved they are done without necessarily taking the python type route. 

See this thread for more context. 

_Originally posted by @jonathan-s in https://github.com/LilyFoote/django-rusty-templates/pull/70#discussion_r1976443279_
            
As a side-effect of this current behaviour, the test cases in rust needs to be written rather awkwardly to cover other match arms.

LilyFoote:
Yeah, I think we should definitely look into this at some point. I think I'd like to get more of the complicated `{% tag %}` implementations done first though.

PR #75, opened 2025-03-02
Extract the variables into python types to get into the correct matcharm from the start

jonathan-s:
Creating this as a draft PR to check-in with the approach. This does seem to solve #71, as the coverage pattern definitely have changed.


LilyFoote:
Ah! I think I misunderstood your proposal in #71. This seems like a good idea to me!

We'll need to handle `Content::HtmlSafe` and `autoescape` of course.

PR #78, opened 2025-04-19
Feat/yesno filter

oliverhaas:
Hello!

I got really interested in the project a few weeks back, and I thought I just have a go at contributing.

I honestly have no experience in rust, so if it gets to a point where my contributions are more a net negative, just let me know.

I don't really understand most parts of the project, and I even have some basic questions. Maybe if I pick one for now: If I'm going to implement several of the filters, do we really need the rust tests in addition to the python tests? The filter usually don't have any fancy logic, and then I personally would be fine with just testing the "blackbox" behavior with the python tests.


LilyFoote:
> Hello!

Hi!

> I got really interested in the project a few weeks back, and I thought I just have a go at contributing.

Awesome!

> I honestly have no experience in rust, so if it gets to a point where my contributions are more a net negative, just let me know.

I'm sure that won't be a problem.

> I don't really understand most parts of the project, and I even have some basic questions. Maybe if I pick one for now: If I'm going to implement several of the filters, do we really need the rust tests in addition to the python tests? The filter usually don't have any fancy logic, and then I personally would be fine with just testing the "blackbox" behavior with the python tests.

For now, I don't really mind whether the tests are written in Rust or in Python, so long as we have (close to) 100% coverage in CI.

src/render/filters.rs
```diff
@@ -319,6 +322,75 @@ impl ResolveFilter for SlugifyFilter {
...
+            
+            // Handle None values
+            if left.to_py(py)?.is_none() {
+                // Return "maybe" if provided, otherwise fallback to "no"
+                let result = if parts.len() >= 3 {
+                    parts[2].to_string()
+                } else {
+                    parts.get(1).unwrap_or(&"no").to_string()
+                };
+                return Ok(Some(Content::String(Cow::Owned(result))));
```
LilyFoote, line 370: I think it would be cleaner to match on the length of `parts` and handle each case explicitly:

```suggestion
                let result = match parts.len() {
                    3 => Cow::Owned(parts[2].to_string()),
                    2 => Cow::Owned(parts[1].to_string()),
                    1 => Cow::Borrowed("no"),
                    0 => todo!("Match Django's behaviour"),
                    _ => todo!("Match Django's behaviour"),
                };
                return Ok(Some(Content::String(result)));
```

src/render/filters.rs
```diff
@@ -319,6 +322,75 @@ impl ResolveFilter for SlugifyFilter {
...
+                        key_at: (0, 0).into(),
+                        object_at: None,
+                    }));
+                }
+            };
+
+            let parts: Vec<&str> = mapping.split(',').collect();
+            
+            // Handle None values
+            if left.to_py(py)?.is_none() {
```
LilyFoote, line 363: I wonder if we should add a new method to `Content` returning an enum with three variants: `None`, `Truthy` and `Falsey`, which we can then match on here.

LilyFoote (changes requested):
This is an awesome start!

I've got a bunch of suggestions for improvement. Please ask if you want more explanation for anything.

Issue #14, opened 2024-10-29
Implement rendering TranslatedText


LilyFoote:
Django handles resolving translations in https://github.com/django/django/blob/main/django/template/base.py#L859-L865.

PR #79, opened 2025-04-27
Implement translations in filter arguments

LilyFoote:
For #14.

### DIFF
--- a/src/render/common.rs
+++ b/src/render/common.rs
@@ -11,6 +11,7 @@ use crate::types::Argument;
 use crate::types::ArgumentType;
 use crate::types::TemplateString;
 use crate::types::Text;
+use crate::types::TranslatedText;
 use crate::types::Variable;
 
 impl Resolve for Variable {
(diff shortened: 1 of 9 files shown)


src/render/common.rs
```diff
@@ -80,6 +81,25 @@ impl Resolve for Text {
...
+        &self,
+        py: Python<'py>,
+        template: TemplateString<'t>,
+        context: &mut Context,
+        _failures: ResolveFailures,
+    ) -> ResolveResult<'t, 'py> {
+        let resolved = Cow::Borrowed(template.content(self.at));
+        let django_translation = py.import("django.utils.translation")?;
+        let get_text = django_translation.getattr("gettext")?;
+        let resolved = get_text.call1((resolved,))?.extract::<String>()?;
```
LilyFoote, line 95: I would prefer to delegate to a Rust implementation here, but there's a whole lot of complexity in how Django finds and loads `.mo` files that we'd have to handle.

LilyFoote:
@jonathan-s I've tagged you for review so you can be more involved. Let me know if you'd prefer not to review things.

Issue #27, opened 2024-10-31
Test safe_join with Windows paths


LilyFoote:
Windows paths behave differently to unix paths so we should ensure `safe_join` handles them appropriately.

Depends on #26.

PR #85, opened 2025-04-30
add OSes to workflow matrix for Python and Rust tests

### LINKED ISSUE #26: Test on Windows
We should test on Windows so we don't miss edge-cases on this platform, for example in path handling.

joshuadavidthomas:
👋 Hiya!

Took a guess at what you meant about testing on Windows in #26 and added an OS key to the workflow matrix for the two test workflows. I went ahead and added macOS as well, but if that's not desired I can always pare it back to Linux and Windows.

closes #26

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -22,8 +22,8 @@ jobs:
         python-version: ["3.10", "3.11", "3.12"]
 
     steps:
-    - name: Configure Git line-endings (Windows)
-      if: runner.os == 'Windows'
+    # on Windows, make sure line-endings are consistent for test suite
+    - if: runner.os == 'Windows'
       run: |
         git config --global core.autocrlf false
         git config --global core.eol lf
@@ -31,15 +31,15 @@ jobs:
     - uses: actions/checkout@v4
 
     - name: Install gettext for translation testing (Ubuntu)
-      if: matrix.os == 'ubuntu-latest'
+      if: runner.os == 'Linux'
       run: sudo apt-get install gettext
 
     - name: Install gettext for translation testing (macOS)
-      if: matrix.os == 'macos-latest'
+      if: runner.os == 'macOS'
       run: brew install gettext
 
     - name: Install gettext for translation testing (Windows)
-      if: matrix.os == 'windows-latest'
+      if: runner.os == 'Windows'
       shell: bash
       run: |
         LATEST_URL=$(curl -s "https://api.github.com/repos/vslavik/gettext-tools-windows/releases/latest" |
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -11,10 +11,11 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
+        os: ["macos-latest", "ubuntu-latest", "windows-latest"]
         python-version: ["3.10", "3.11", "3.12"]
 
     steps:
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -59,11 +59,12 @@ jobs:
         .venv/bin/pip install pytest-cov
     - name: Install dependencies (Windows)
       if: matrix.os == 'windows-latest'
+      shell: bash
       run: |
         python -m venv .venv
-        .venv\Scripts\python -m pip install --upgrade pip
-        .venv\Scripts\pip install -r requirements.txt
-        .venv\Scripts\pip install pytest-cov
+        .venv/Scripts/python -m pip install --upgrade pip
+        .venv/Scripts/pip install -r requirements.txt
+        .venv/Scripts/pip install pytest-cov
     - name: Install Django Rusty Templates (Unix)
       if: matrix.os != 'windows-latest'
       run: |
@@ -82,7 +83,7 @@ jobs:
       run: .venv/bin/python -m django compilemessages
     - name: Build translation files (Windows)
       if: matrix.os == 'windows-latest'
-      run: .venv\Scripts\python -m django compilemessages
+      run: .venv/Scripts/python -m django compilemessages
     - name: Lint test files (Unix)
       if: matrix.os != 'windows-latest'
       run: |
@@ -90,7 +91,7 @@ jobs:
     - name: Lint test files (Windows)
       if: matrix.os == 'windows-latest'
       run: |
-        .venv\Scripts\ruff check
+        .venv/Scripts/ruff check
     - name: Test with pytest (Unix)
       if: matrix.os != 'windows-latest'
       run: |
@@ -103,7 +104,7 @@ jobs:
       run: |
         cargo llvm-cov show-env --export-prefix > llvm-cov-env.sh
         source llvm-cov-env.sh
-        .venv\Scripts\pytest --cov --cov-report=xml
+        .venv/Scripts/pytest --cov --cov-report=xml
     - name: Get rust coverage report (Unix)
       if: runner.os != 'Windows'
       run: |
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -32,11 +32,6 @@ jobs:
     - name: Install cargo-llvm-cov
       uses: taiki-e/install-action@cargo-llvm-cov
 
-    - name: Setup LLVM coverage environment
-      run: |
-        cargo llvm-cov show-env --export-prefix > llvm-cov-env.sh
-        echo "source $(pwd)/llvm-cov-env.sh" >> $GITHUB_ENV
-
     - name: Install dependencies
       run: |
         python -m venv .venv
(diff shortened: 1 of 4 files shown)


LilyFoote:
Hey @joshuadavidthomas, thanks for making a start with this!

To get this PR merged, I think we should get the tests running on each OS, but we don't need all tests passing - we can skip them based on the OS instead and fix them in follow-up PRs. If you feel motivated to fix them in this PR that's good too.

joshuadavidthomas:
@LilyFoote Sounds like a plan! The problems look simple enough (at least at first glance), I don’t mind trying to fix in this PR.

joshuadavidthomas:
Ah damn, you'll have to approve any change for the CI workflows to run. That's going to be a major pain as I try to pick away at this. I'll just do it on my fork for now and bring over the completed fixes all at once.

LilyFoote:
Hmm, I don't appear to be able to approve this latest change. Is there a syntax error or something?

joshuadavidthomas:
> Hmm, I don't appear to be able to approve this latest change. Is there a syntax error or something?

Maybe! I wouldn't worry about it for now -- I've got a PR against my fork I'm working through this on that seems to be working: https://github.com/joshuadavidthomas/django-rusty-templates/pull/7

Only thing that doesn't work is the Codecov action, but minor detail. I've got macOS and Ubuntu done, just hitting my head against Windows at the moment. Right now I've got the different steps intermixed in the single job, but it's getting a bit hard to read (IMO!). Would you mind separate jobs for at least Windows and the Unix based platforms? Or just keep them all together?

LilyFoote:
> > Hmm, I don't appear to be able to approve this latest change. Is there a syntax error or something?
> 
> Maybe! I wouldn't worry about it for now -- I've got a PR against my fork I'm working through this on that seems to be working: [joshuadavidthomas#7](https://github.com/joshuadavidthomas/django-rusty-templates/pull/7)

It only affected one commit, so I'm not worried about it now.

> Only thing that doesn't work is the Codecov action, but minor detail. I've got macOS and Ubuntu done, just hitting my head against Windows at the moment. Right now I've got the different steps intermixed in the single job, but it's getting a bit hard to read (IMO!). Would you mind separate jobs for at least Windows and the Unix based platforms? Or just keep them all together?

Whatever you think is most maintainable is good.

joshuadavidthomas:
@LilyFoote Okay, all greens across the board on CI. Couple of notes:

- Relying on [this repo](https://github.com/vslavik/gettext-tools-windows) to install GNU gettext on Windows, just to get something working for now. It looks relatively active though, with a release from January this year. Can't vouch for it beyond that, so exploring other options later on may be desired.
- Went ahead and activated the virtualenv (via two separate steps, one for Linux/macOS and one for Windows) so the rest of the commands could be the same for all the platforms
- Using bash as the default shell, to get around Windows and calling `source`.
- Changed the `llvm-cov` environment to use a temp shell script to get around shell piping issues on macOS using the `<( ... )` syntax.
- Adjusted one Python test and a handful of Rust tests to account for the differences in path separators on Windows. I went with `#[cfg(not(windows))]` as opposed to `#[cfg(unix)]`, but if you want I can change that.
- Ignored one test on Windows because it looked like a can of worms I didn't necessarily want to open right now. 😄

.github/workflows/python.yml
```diff
@@ -31,29 +56,52 @@ jobs:
...
-        .venv/bin/ruff check
+        ruff check
+
     - name: Test with pytest
       run: |
-        source <(cargo llvm-cov show-env --export-prefix)
-        .venv/bin/pytest --cov --cov-report=xml
+        cargo llvm-cov show-env --export-prefix > llvm-cov-env.sh
+        source llvm-cov-env.sh
+        pytest --cov --cov-report=xml -v
```
joshuadavidthomas, line 97: Ah, the `-v` verbose flag can be removed, this was to see a bit more output from the pytest failure

.github/workflows/python.yml
```diff
@@ -11,18 +11,43 @@ on:
...
+      if: runner.os == 'Linux'
       run: sudo apt-get install gettext
 
+    - name: Install gettext for translation testing (macOS)
+      if: runner.os == 'macOS'
+      run: brew install gettext
+
+    - name: Install gettext for translation testing (Windows)
+      if: runner.os == 'Windows'
+      shell: bash
```
joshuadavidthomas, line 43: default is bash now, this can be removed

LilyFoote:
>     * Relying on [this repo](https://github.com/vslavik/gettext-tools-windows) to install GNU gettext on Windows, just to get something working for now. It looks relatively active though, with a release from January this year. Can't vouch for it beyond that, so exploring other options later on may be desired.

The [Django docs recommend](https://docs.djangoproject.com/en/5.2/topics/i18n/translation/#gettext-on-windows) a different source of [binary installers](https://mlocati.github.io/articles/gettext-iconv-windows.html). I'm not sure if there's any reason to prefer one over the other.

>     * Changed the `llvm-cov` environment to use a temp shell script to get around shell piping issues on macOS using the `<( ... )` syntax.

Definitely out of scope for this PR, but I've been pondering the possibility of adding a [Justfile](https://just.systems/man/en/) with a way to run coverage locally. I tried it once but never got it working because of issues with the environment variables.

>     * Ignored one test on Windows because it looked like a can of worms I didn't necessarily want to open right now. 😄

Fair! I think this would be in scope for #27.

tests/test_django.py
```diff
@@ -20,15 +21,16 @@ def test_parse_error():
         get_template("parse_error.txt", using="rusty")
 
     template_dir = Path("tests/templates").absolute()
+    path_separator = '/' if os.name != 'nt' else '\\'
```
LilyFoote, line 24: We can use [`os.sep`](https://docs.python.org/3/library/os.html#os.sep) instead:

```suggestion
```

tests/test_django.py
```diff
@@ -20,15 +21,16 @@ def test_parse_error():
...
   × Empty variable tag
-   ╭─[%s/parse_error.txt:1:28]
+   ╭─[%s%sparse_error.txt:1:28]
  1 │ This is an empty variable: {{ }}
    ·                            ──┬──
    ·                              ╰── here
    ╰────
 """
-    assert str(excinfo.value) == expected % template_dir
+    assert str(excinfo.value) == expected % (template_dir, path_separator)
```
LilyFoote, line 33:
```suggestion
    assert str(excinfo.value) == expected % (template_dir, os.sep)
```

LilyFoote (approved):
Amazing work, thanks!

I see one place we can tidy up a bit and I see you've noticed a few others yourself, but I'm happy to land this when you are.

tests/test_django.py
joshuadavidthomas: TIL! They have everything in the stdlib, what will they think of next?!

joshuadavidthomas:
> The [Django docs recommend](https://docs.djangoproject.com/en/5.2/topics/i18n/translation/#gettext-on-windows) a different source of [binary installers](https://mlocati.github.io/articles/gettext-iconv-windows.html). I'm not sure if there's any reason to prefer one over the other.

Oh, I just landed on the one I did because it was the first search result for "gettext Windows GitHub Actions". Differences are probably negligible, but I would lean towards using what's recommended by Django if only because it's in the docs for a reason. (Whether I know that reason or not is irrelevant to me. 😄) Should be a simple swap of the URLs in that step.

> Definitely out of scope for this PR, but I've been pondering the possibility of adding a [Justfile](https://just.systems/man/en/) with a way to run coverage locally. I tried it once but never got it working because of issues with the environment variables.

OH yes, I'm a big fan of just and [use] [Justfiles] [in] [all]() [of]() [my]() [projects](). I've tried a bunch of the patterns of how to layout Justfiles/use just and have liked where I've landed recently with modules to group commands. Even just putting one line simple commands behind a just commands helps improve the dev experience, since you can swap out the underlying command while keeping the CLI interface consistent. (E.g., swapping from pip-tools to uv for dependency management.) Big +1 from me.

Also interested in getting coverage on [my Django/Rust project], but I'm a relatively new to the Rust ecosystem so I haven't had the time to dig in to how to actually go about setting it up. I'm definitely going to be ~stealing~ borrowing some ideas 👀.

[use]: https://github.com/joshuadavidthomas/django-bird-autoconf/blob/main/Justfile
[Justfiles]: https://github.com/joshuadavidthomas/django-github-app/blob/main/Justfile
[in]: https://github.com/joshuadavidthomas/django-bird/blob/main/Justfile
[all]: https://github.com/joshuadavidthomas/django-language-server/blob/main/Justfile
[of]: https://github.com/joshuadavidthomas/django-lazy-gdal/blob/main/Justfile
[my]: https://github.com/joshuadavidthomas/llm-uv-tool/blob/main/Justfile
[projects]: https://github.com/westerveltco/django-twc-project/blob/main/examples/default/Justfile
[my Django/Rust project]: https://github.com/joshuadavidthomas/django-language-server

PR #89, opened 2025-05-09
Hook-all-loaders

KarimTayie:


src/template.rs
```diff
@@ -112,12 +112,115 @@ pub mod django_rusty_templates {
...
+                    &loader_str.extract::<String>()?,
+                    Vec::new(),
+                    HashMap::new(),
+                );
+            }
+
+            // Try as tuple (loader, args)
+            if let Ok(loader_tuple) = ld.downcast::<PyTuple>() {
+                if let Ok((loader_obj, args)) =
+                    loader_tuple.extract::<(Bound<'_, PyString>, Bound<'_, PyAny>)>()
```
LilyFoote, line 142: The [Python code we're re-implementing](https://github.com/django/django/blob/42ab99309d347f617d60751c2e8d627fb2963049/django/template/engine.py#L142-L143) does:

```python
        if isinstance(loader, (tuple, list)):
            loader, *args = loader
```

but the Rust code here only handles the case where `loader` has two elements. I think we should write a helper function:

```rust
fn unpack(loader: Bound<'py, PyAny>) -> PyResult<(Bound<'py, PyString>, Vec<Bound<'py, PyAny>>)> {
    let mut items = loader.try_iter()?;
    let loader_path = items.next().downcast()?;
    Ok((loader_path, items.collect()))
}
```

src/template.rs
```diff
@@ -112,12 +112,115 @@ pub mod django_rusty_templates {
...
+                            .collect::<Result<HashMap<String, String>, PyErr>>()?;
+
+                        return Self::map_loader(py, &loader_path, Vec::new(), args_hashmap);
+                    } else {
+                        return Err(PyErr::new::<pyo3::exceptions::PyTypeError, _>(format!(
+                            "Unsupported args type: {:?}",
+                            args
+                        )));
+                    }
+                }
```
LilyFoote, line 182: With `unpack` we can do something like:

```suggestion
                let (loader, args) = unpack(loader)?;
                return Self::map_loader(py, loader.to_str()?, args);
```

and we can move the handling of the contents of `args` into `map_loader` where we know which loader we're working with.

src/template.rs
```diff
@@ -112,12 +112,115 @@ pub mod django_rusty_templates {
...
+                "Unsupported loader type: {:?}",
+                ld
+            )))
+        }
+
+        fn map_loader(
+            py: Python<'_>,
+            loader_path: &str,
+            args_vec: Vec<String>,
+            args_hashmap: HashMap<String, String>,
```
LilyFoote, line 195: With the `unpack` refactor:

```suggestion
            args: Vec<Bound<'_, PyAny>>,
```

tests/test_engine.py
```diff
@@ -53,5 +52,83 @@ def test_pathlib_dirs():
...
+def test_cached_loader_priority():
+    loaders = [
+        (
+            "django.template.loaders.cached.Loader",
+            [
+                "django.template.loaders.filesystem.Loader",
+                "django.template.loaders.app_directories.Loader",
+            ],
+        ),
+    ]
```
LilyFoote, line 93: We also need to be able to handle this:

```python
    loaders = [
        (
            "django.template.loaders.cached.Loader",
            [
                ("django.template.loaders.filesystem.Loader", ["tests", "app"]),
                "django.template.loaders.app_directories.Loader",
            ],
        ),
    ]
```

LilyFoote (changes requested):
Thanks for making a start with this!

I've left a bunch of feedback on things I think we can improve.

KarimTayie:
> Thanks for making a start with this!
> 
> I've left a bunch of feedback on things I think we can improve.

Thank you for the thorough review and detailed feedback! I really appreciate the time and effort you put into it. I’ll start addressing the comments as soon as I have the time.

PR #91, opened 2025-05-22
add center filter

moshe742:


tests/filters/test_center.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+
+
```
LilyAcorn, line 3:
```suggestion
```

LilyAcorn:
I've made a change on `main` that I think resolves the codecov issue, so the tests should work if you merge or rebase `main`.

tests/filters/test_center.py
moshe742: Done

LilyAcorn:
@moshe742 Sorry for the huge delay! I *think* I've finally fixed the coverage issue, so can you please sync with `main` to trigger the builds again?

src/parse.rs
```diff
@@ -105,6 +106,10 @@ impl Filter {
                 Some(right) => return Err(unexpected_argument("capfirst", right)),
                 None => FilterType::Capfirst(CapfirstFilter),
             },
+            "center" => match right {
+                Some(right) => FilterType::Center(CenterFilter::new(right)),
+                None => return Err(ParseError::MissingArgument { at: at.into() }),
```
LilyFirefly, line 111: We need a test for this line.

src/render/filters.rs
```diff
@@ -164,6 +167,47 @@ impl ResolveFilter for CapfirstFilter {
...
+        context: &mut Context,
+    ) -> ResolveResult<'t, 'py> {
+        let left: usize;
+        let right: usize;
+        let arg_size: usize;
+        let content = match variable {
+            Some(content) => {
+                content.render(context)?.into_owned()
+            },
+            None => return Ok("".as_content()),
```
LilyFirefly, line 185: This needs a test.

src/render/filters.rs
```diff
@@ -164,6 +167,47 @@ impl ResolveFilter for CapfirstFilter {
...
+            Some(arg_int) => {
+                arg_size = arg_int.to_usize().unwrap();
+                if arg_size <= content.len() {
+                    return Ok(content.into_content());
+                }
+                right = (arg_size - content.len()) / 2;
+                left = arg_size - content.len() - right;
+            },
+            None => {
+                return Ok("".as_content());
```
LilyFirefly: This needs a test.

src/render/filters.rs
```diff
@@ -164,6 +167,47 @@ impl ResolveFilter for CapfirstFilter {
...
+            Some(content) => {
+                content.render(context)?.into_owned()
+            },
+            None => return Ok("".as_content()),
+        };
+        let arg = self
+            .argument
+            .resolve(py, template, context, ResolveFailures::Raise)?
+            .expect("missing argument in context should already have raised");
+        let arg_int = arg.to_bigint();
```
LilyFirefly: We don't need to convert to a big int - a Rust string can't be longer than `usize::MAX`. If the `arg` can't be turned into a `usize`, we can return an error.

src/render/filters.rs
```diff
@@ -164,6 +167,47 @@ impl ResolveFilter for CapfirstFilter {
...
+        match arg_int {
+            Some(arg_int) => {
+                arg_size = arg_int.to_usize().unwrap();
```
LilyFirefly: As noted above, I'd like us to return an error if the `arg` can't be turned into a `usize` - not panic, which `unwrap()` would do.

src/render/filters.rs
```diff
@@ -164,6 +167,47 @@ impl ResolveFilter for CapfirstFilter {
...
+        py: Python<'py>,
+        template: TemplateString<'t>,
+        context: &mut Context,
+    ) -> ResolveResult<'t, 'py> {
+        let left: usize;
+        let right: usize;
+        let arg_size: usize;
+        let content = match variable {
+            Some(content) => {
+                content.render(context)?.into_owned()
```
LilyFirefly: I don't think we should need to convert to an `owned` string here.

src/render/filters.rs
```diff
@@ -164,6 +167,47 @@ impl ResolveFilter for CapfirstFilter {
...
+                right = (arg_size - content.len()) / 2;
+                left = arg_size - content.len() - right;
+            },
+            None => {
+                return Ok("".as_content());
+            }
+        }
+        let string_left = " ".repeat(left);
+        let string_right = " ".repeat(right);
+        Ok((string_left + &content + &string_right).into_content())
```
LilyFirefly: I think we can slightly optimise here by using [`with_capacity`](https://doc.rust-lang.org/std/string/struct.String.html#method.with_capacity):

```rust
       let mut centered = String::with_capacity(arg_size);
       centered.push_str(&" ".repeat(left));
       centered.push_str(&content);
       centered.push_str(&" ".repeat(right);
       Ok(centered.into_content())
```

LilyAcorn (commented):
Thanks so much for the patience waiting for this review!

This looks like a really good start. I have a few suggestions that I think will improve it further.

src/render/filters.rs
moshe742: Done.
I didn't do the below suggestion since the result of the rust implementation is not the same as the Django implementation.
So, do we want to change the result from what is documented (and then the below implementation is probably better) or not?

LilyFirefly: Interesting! Can you elaborate on what the difference is? We probably do want to match Django's behaviour here, but I'd like to understand better before giving a final opinion.

moshe742: When I use the below code I get the following:
assertion `left == right` failed
  left: "    django     "
 right: "     django    "

---------------------------
When left is the output from the rust code and the right is the test string literal

So, if there is odd number of spaces to insert django puts the larger number at the start of the string, rust is doing it at the end.

If my explanation is not clear ask again with what is not clear so I can try to do better

LilyFirefly: Ok, let's go with the Django behaviour then.

src/parse.rs
moshe742: Done

src/render/filters.rs
moshe742: Done

src/render/filters.rs
moshe742: Done

src/render/filters.rs
moshe742: Added a to_usize on render/types and used it for this, since I assume it will be used later on other filters too.
I didn't return an error (returns a OK, since I wasn't able to understand how to return an error, so you might want to change that (or if you can explain to me how to do it, even better)

src/render/filters.rs
moshe742: Done, as noted above, return Ok since I don't understand how to return an Err (with a reason for the problem)

src/render/filters.rs
moshe742: I tried to remove it but then I need to use as_content on line 198 and then I get an error that I try to return a value referencing data owned by the current function since content is borrowed by as_content.
I don't understand how to fix this, so if you can help it will be appreciated

src/render/filters.rs
LilyFirefly: Actually, investigating this even more, it turns out that [Python's `str.center` method is **weird**](https://github.com/python/cpython/issues/67812):

```python
>>> "odd".center(6)
' odd  '
>>> "even".center(7)
'  even '
```

And since Django just delegates to `str.center`, we also inherit this really unintuitive behaviour.

Frustratingly, f-strings behave consistently and match Rust's behaviour:

```python
>>> f"{'odd':^6}"
' odd  '
>>> f"{'even':^7}"
' even  '
```

But since we're aiming for compatibility with Django, we're stuck with the weird behaviour.

src/render/filters.rs
```diff
@@ -2,12 +2,12 @@ use std::borrow::Cow;
 use std::sync::LazyLock;
 
 use html_escape::encode_quoted_attribute_to_string;
-use pyo3::prelude::*;
+use pyo3::{prelude::*, IntoPyObjectExt};
```
LilyFirefly: I don't think `IntoPyObjectExt` is used currently.

```suggestion
use pyo3::prelude::*;
```

src/render/filters.rs
```diff
@@ -164,6 +165,53 @@ impl ResolveFilter for CapfirstFilter {
...
+            .resolve(py, template, context, ResolveFailures::Raise)?
+            .expect("missing argument in context should already have raised");
+        let arg_size = arg.to_usize();
+        let argument_size;
+
+        match arg_size {
+            Some(size) => {
+                argument_size = size;
+                if size <= content.len() {
+                    return Ok(content.into_content());
```
LilyFirefly: If we do this, we won't need `content.into_owned()` above.

```suggestion
                    return Ok(Some(Content::String(ContentString::String(content))));
```

src/render/types.rs
```diff
@@ -108,6 +109,30 @@ impl<'t, 'py> Content<'t, 'py> {
...
+                    let int = PyType::new::<PyInt>(left.py());
+                    match int.call1((left,)) {
+                        Ok(left) => Some(
+                            left.extract::<usize>()
+                                .expect("Python integers are usize compatible"),
+                        ),
+                        Err(_) => None,
+                    }
+                }
+            },
```
LilyFirefly: We should have tests that cover these branches.

src/render/types.rs
```diff
@@ -108,6 +109,30 @@ impl<'t, 'py> Content<'t, 'py> {
         })
     }
 
+    pub fn to_usize(&self) -> Option<usize> {
```
LilyFirefly: This would be cleaner if we return a `Result` instead of an `Option`.

```suggestion
    pub fn to_usize(&self) -> Result<usize, PyRenderError> {
```

`PyRenderError` can be found in `src/error.rs` and is either a `PyErr` or a `RenderError`. If there isn't an appropriate `PyErr` to return, we can define a new `RenderError` variant and return that instead.

src/render/types.rs
```diff
@@ -108,6 +109,30 @@ impl<'t, 'py> Content<'t, 'py> {
...
+            },
+            Self::Float(left) => left.trunc().to_usize(),
+            Self::Py(left) => match left.extract::<usize>() {
+                Ok(left) => Some(left),
+                Err(_) => {
+                    let int = PyType::new::<PyInt>(left.py());
+                    match int.call1((left,)) {
+                        Ok(left) => Some(
+                            left.extract::<usize>()
+                                .expect("Python integers are usize compatible"),
```
LilyFirefly: This isn't true. A Python integer can be negative or bigger than `usize::MAX`. We'll need to return an `Err` here instead.

src/render/filters.rs
```diff
@@ -164,6 +165,53 @@ impl ResolveFilter for CapfirstFilter {
...
+            Some(content) => {
+                content.render(context)?.into_owned()
+            },
+            None => return Ok("".as_content()),
+        };
+        let arg = self
+            .argument
+            .resolve(py, template, context, ResolveFailures::Raise)?
+            .expect("missing argument in context should already have raised");
+        let arg_size = arg.to_usize();
```
LilyFirefly: With the change to the signature of `to_usize` suggested below:

```suggestion
        let arg_size = arg.to_usize()?;
```

This will also mean that `arg_size` is guaranteed to be a `usize` at this point, allowing the code below to be simplified (remove the `match` and the `None` branch).

src/render/filters.rs
LilyFirefly: I've opened a bug report in Django: https://code.djangoproject.com/ticket/36519

But we won't be able to change the behaviour here until that lands.

src/render/types.rs
LilyFirefly: We also need to handle negative numbers carefully - Django's `center` filter accepts them (treats them essentially as zero), so we need to too.

src/render/filters.rs
moshe742: Done, thanks!

src/render/types.rs
```diff
@@ -108,6 +110,44 @@ impl<'t, 'py> Content<'t, 'py> {
         })
     }
 
+    pub fn to_usize(&self) -> Result<usize, PyRenderError> {
+        match self {
+            Self::Int(left) => {
+                match left.sign() {
+                    Sign::Minus => Ok(0),
+                    Sign::Plus => {
+                        let result = left.to_usize().expect("Int can always be converted to usize");
```
LilyFirefly: This is not true - `BigInt` can be bigger than `usize::MAX`. We should return an error here instead of panicking.

src/render/types.rs
```diff
@@ -108,6 +110,44 @@ impl<'t, 'py> Content<'t, 'py> {
...
+                        Ok(result)
+                    },
+                    Sign::NoSign => {
```
LilyFirefly: Is it only `0` that has `NoSign`? If so, we can merge this case into the `Sign::Minus` case.

src/render/types.rs
```diff
@@ -108,6 +110,44 @@ impl<'t, 'py> Content<'t, 'py> {
...
+                }
+            },
+            Self::String(left) => match left.as_raw().parse::<usize>() {
+                Ok(left) => Ok(left),
+                Err(_) => Err(PyRenderError::PyErr(
+                    // TODO: check the error
+                    PyValueError::new_err(format!("invalid literal for int() with base 10: '{}'", left.as_raw(),
+                )))),
+            },
+            Self::Float(left) => Ok(left.trunc().to_usize().expect("f64 can always be converted to usize")),
```
LilyFirefly: This isn't true. Consider negative floats, `inf` and `nan`.

src/render/types.rs
```diff
@@ -108,6 +110,44 @@ impl<'t, 'py> Content<'t, 'py> {
...
+                        Ok(result)
+                    },
+                }
+            },
+            Self::String(left) => match left.as_raw().parse::<usize>() {
+                Ok(left) => Ok(left),
+                Err(_) => Err(PyRenderError::PyErr(
+                    // TODO: check the error
+                    PyValueError::new_err(format!("invalid literal for int() with base 10: '{}'", left.as_raw(),
+                )))),
```
LilyFirefly: Thinking about this some more, I'd prefer we raise a custom `RenderError` here so we can provide a better debugging experience than Django does.

Something like:

```rust
#[derive(Error, Debug, Diagnostic, PartialEq, Eq)]
pub enum RenderError {
    #[error(Couldn't convert argument {argument} to integer)]
    InvalidArgumentInteger {
        argument: String,
        #[label("argument"}]
        argument_at: SourceSpan,
    },
    // other variants
}
```

```suggestion
                Err(_) => Err(RenderError::InvalidArgumentInteger {
                    argument: left,
                    argument_at: argument.at.into(),
                }.into()),
```

I think you'll need to refactor to bring `argument.at` into scope somehow. Perhaps simplest is to move this logic back into the `center` filter for now. We can always extract it again later if we need it for another filter.

src/render/types.rs
```diff
@@ -108,6 +110,44 @@ impl<'t, 'py> Content<'t, 'py> {
...
+            Self::Float(left) => Ok(left.trunc().to_usize().expect("f64 can always be converted to usize")),
+            Self::Py(left) => match left.extract::<usize>() {
+                Ok(left) => Ok(left),
+                Err(_) => {
+                    let int = PyType::new::<PyInt>(left.py());
+                    match int.call1((left,)) {
+                        Ok(left) => Ok(left.extract::<usize>()?),
+                        Err(_) => Err(PyRenderError::PyErr(
+                            PyValueError::new_err(format!("invalid literal for int() with base 10: '{}'", left),
+                        ))),
```
LilyFirefly: Similarly, I'd like us to use `RenderError::InvalidArgumentInteger` here too.

LilyAcorn (commented):
Thanks again for persisting with this! While there's still more we can improve, I really feel like we're making great progress!

src/render/types.rs
moshe742: OK, dealt with negative floats, but inf and nan are considered strings, so I might be able to give a good error message, but it's not going to the float section at all, it goes to the string section

LilyFirefly: See a6f0eb10ddf160f9a10075657c63554b626802d0 for a trick to get `inf` and `-inf`. We might not need to special-case `nan` - I can't remember there being a trick for it.

src/render/types.rs
moshe742: When I use this I get for some reason the error:django.template.base.VariableDoesNotExist
but with the right string, is this what we want for now (I assume that once we have the full implementation we would want to have the right error coming from the template)?

src/render/filters.rs
```diff
@@ -164,6 +170,102 @@ impl ResolveFilter for CapfirstFilter {
...
+        py: Python<'py>,
+        template: TemplateString<'t>,
+        context: &mut Context,
+    ) -> ResolveResult<'t, 'py> {
+        let left: usize;
+        let right: usize;
+        let content = match variable {
+            Some(content) => {
+                content.render(context)?
+            },
```
```suggestion
            Some(content) => content.render(context)?,
```

src/render/filters.rs
```diff
@@ -164,6 +170,102 @@ impl ResolveFilter for CapfirstFilter {
...
+                    },
+                }
+            },
+            Content::String(left) => match left.as_raw().parse::<usize>() {
+                Ok(left) => Ok(left),
+                Err(_) => Err(PyRenderError::PyErr(
+                    // TODO: check the error
+                    PyValueError::new_err(format!("invalid literal for int() with base 10: '{}'", left.as_raw(),
+                )))),
+            },
```
LilyFirefly: We should check how Django handles a string that can be converted to a negative integer.

src/render/filters.rs
```diff
@@ -164,6 +170,102 @@ impl ResolveFilter for CapfirstFilter {
...
+                            }.into())
+                        },
+                    }
+                }
+            },
+        };
+        let size = arg_size?;
+
+        if size <= content.len() {
+            return Ok(Some(Content::String(ContentString::String((content)))));
```
```suggestion
            return Ok(Some(Content::String(ContentString::String(content))));
```

src/render/filters.rs
```diff
@@ -164,6 +170,102 @@ impl ResolveFilter for CapfirstFilter {
...
+                Ok(left) => Ok(left),
+                Err(_) => Err(PyRenderError::PyErr(
+                    // TODO: check the error
+                    PyValueError::new_err(format!("invalid literal for int() with base 10: '{}'", left.as_raw(),
+                )))),
+            },
+            Content::Float(left) => {
+                let result = left.trunc();
+                if result <= 0f64 {
+                    return Ok(content.to_string().into_content());
```
LilyFirefly: We can avoid the (minor) overhead of `to_string()`:

```suggestion
                    return Ok(Some(Content::String(ContentString::String(content))));
```

src/render/types.rs
```diff
@@ -108,6 +110,55 @@ impl<'t, 'py> Content<'t, 'py> {
...
+    //                     Ok(left) => Ok(left.extract::<usize>()?),
+    //                     Err(_) => Err(PyRenderError::PyErr(
+    //                         PyValueError::new_err(format!("literal for int() with base 10: '{}'", left),
+    //                     ))),
+    //                 }
+    //             }
+    //         },
+    //     }
+    // }
+
```
LilyFirefly: This can now be deleted.

src/render/types.rs
```diff
@@ -2,12 +2,14 @@ use std::borrow::Cow;
...
-use num_bigint::{BigInt, ToBigInt};
-use pyo3::exceptions::PyAttributeError;
+use num_bigint::{BigInt, Sign, ToBigInt};
+use num_traits::{Float, ToPrimitive};
+use pyo3::exceptions::{PyAttributeError, PyValueError};
 use pyo3::intern;
 use pyo3::prelude::*;
 use pyo3::types::{PyInt, PyString, PyType};
 
+use crate::error::PyRenderError;
```
LilyFirefly: We can return these imports to what they were.

src/error.rs
```diff
@@ -21,6 +22,12 @@ impl PyRenderError {
 
 #[derive(Error, Debug, Diagnostic, PartialEq, Eq)]
 pub enum RenderError {
+    #[error("Couldn't convert argument {argument} to integer")]
```
LilyFirefly: Looking at the tests, I think this will be clearer:

```suggestion
    #[error("Couldn't convert argument ({argument}) to integer")]
```

src/error.rs
```diff
@@ -21,6 +22,12 @@ impl PyRenderError {
...
+    InvalidArgumentInteger {
```
LilyFirefly, line 25: I've realised we currently turn all `RenderError` variants into `VariableDoesNotExist`: https://github.com/LilyAcorn/django-rusty-templates/blob/0922d7d268a4056308e287342feb1000a520b271/src/template.rs#L289-L292

We should update this to match on `err` and return a `PyValueError` for the `InvalidArugmentInteger` variant and `VariableDoesNotExist` for the other two variants.

LilyAcorn (commented):
Sorry this PR is turning into quite the marathon! We're definitely getting there!

src/render/types.rs
LilyFirefly: Ah, I missed this comment - here's the answer: https://github.com/LilyAcorn/django-rusty-templates/pull/91#discussion_r2233998429

LilyAcorn:
Thanks @moshe742! I decided to add the final tweaks myself in #110. I really appreciate the effort you put into this!

PR #100, opened 2025-07-26
Implement the 'first' filter

ipmb:
Full disclosure upfront, this was generated with Claude Code. I'm excited about this project and some of this porting work felt like a good fit for an LLM. If this isn't welcome, feel free to close it. I'm happy to iterate on it if there is feedback.

---

### Implementation Details

The filter returns the first item of a sequence (list, tuple, string) or empty string for empty sequences. It matches Django's behavior, including error handling.

### Django References

- Filter implementation: [django/template/defaultfilters.py#L599-L605](https://github.com/django/django/blob/5.2.4/django/template/defaultfilters.py#L599-L605)
- Django tests: [tests/template_tests/filter_tests/test_first.py](https://github.com/django/django/blob/5.2.4/tests/template_tests/filter_tests/test_first.py)

### Changes

- Added `FirstFilter` struct and enum variant
- Implemented parsing for the 'first' filter (no arguments)
- Implemented `ResolveFilter` trait with Django-compatible behavior
  - Returns first item for sequences
  - Returns empty string for empty sequences
  - Raises `TypeError` for `None` and non-subscriptable types
  - Uses exact Django error messages
- Added comprehensive test suite porting all Django test cases plus additional edge cases

### Testing

The implementation handles:
- Lists, tuples, strings (including empty)
- Falsy values (0, False) correctly returned
- Autoescape on/off behavior
- Safe strings preservation
- Type errors for None and integers


src/render/filters.rs
```diff
@@ -234,6 +235,70 @@ impl ResolveFilter for ExternalFilter {
...
+                            Err(e) => {
+                                // Check if it's an IndexError
+                                if e.is_instance_of::<pyo3::exceptions::PyIndexError>(py) {
+                                    // IndexError -> return empty string
+                                    "".as_content()
+                                } else {
+                                    // Other errors (like TypeError) -> propagate
+                                    return Err(e.into());
+                                }
+                            }
```
LilyFirefly: I'd import `PyIndexError` at the top of the module. We can also reduce the verbosity by inlining the `isinstance` check into the `match`:

```suggestion
                            Err(e) if e.is_instance_of::<PyIndexError>(py) => "".as_content(),
                            Err(e) => return Err(e.into()),
```

src/render/filters.rs
```diff
@@ -234,6 +235,70 @@ impl ResolveFilter for ExternalFilter {
...
+                    Content::String(s) => {
+                        // For strings, get the first character
+                        let s_raw = s.as_raw();
+                        if s_raw.is_empty() {
+                            "".as_content()
+                        } else {
+                            // Get the first character
+                            let first_char = s_raw.chars().next().unwrap();
+                            first_char.to_string().into_content()
+                        }
```
LilyFirefly: I think we can be more efficient here with the [`str.split` method](https://doc.rust-lang.org/std/primitive.str.html#method.split), which returns an iterator of `&str` slices, avoiding allocating a `char` or a `String`:

```suggestion
                        // Skip the empty string always present at the start after splitting
                        let chars = s.split("").skip(1);
                        match chars.next() {
                            None => "".as_content(),
                            c => c.as_content(),
                        }
```

src/render/filters.rs
```diff
@@ -234,6 +235,70 @@ impl ResolveFilter for ExternalFilter {
...
+        &self,
+        variable: Option<Content<'t, 'py>>,
+        py: Python<'py>,
+        _template: TemplateString<'t>,
+        _context: &mut Context,
+    ) -> ResolveResult<'t, 'py> {
+        let content = match variable {
+            Some(content) => {
+                // Try to get the first item from the content
+                match content {
```
LilyFirefly: We can save a lot of horizontal space by moving the `match content` to top level:

```rust
            Some(content) => content,
            None => return Ok("".as_content()),
        };
        let content = match content {
```

src/render/filters.rs
```diff
@@ -234,6 +235,70 @@ impl ResolveFilter for ExternalFilter {
...
+                        .into());
+                    }
+                    Content::Float(_) => {
+                        // Floats are not sequences, should raise TypeError
+                        // Match Django's behavior exactly
+                        return Err(pyo3::exceptions::PyTypeError::new_err(
+                            "'float' object is not subscriptable",
+                        )
+                        .into());
+                    }
```
LilyFirefly: One of the goals of Django Rusty Templates is to provide better error messages than Django often does, with the help of the `miette` crate. In this case, we could define custom `error::RenderError` variants and then convert them into `PyTypeError` in `template::Template::_render`, like we already do with `VariableDoesNotExist`:

https://github.com/LilyAcorn/django-rusty-templates/blob/0922d7d268a4056308e287342feb1000a520b271/src/template.rs#L289-L292

tests/test_filters_first.py
LilyFirefly, line 1: I prefer to test Django's implementation first and then Rust's. This means that we always get a failure if Django's behaviour is not what we expect and we can focus on that before worrying about our own implementation.

tests/test_filters_first.py
```diff
@@ -0,0 +1,243 @@
...
+
+        rusty_result = rusty_engine.from_string(template).render(context)
+        django_result = django_engine.from_string(template).render(context)
+
+        assert rusty_result == "&lt;b&gt;bold&lt;/b&gt;"
+        assert django_result == "&lt;b&gt;bold&lt;/b&gt;"
+
+    def test_first_escaping_with_safe(self):
+        """Test first filter with both escaped and safe strings (Django's first01)."""
+        from django.utils.safestring import mark_safe
```
LilyFirefly: We can move this to module scope. (And in the other tests below.)

tests/test_filters_first.py
```diff
@@ -0,0 +1,243 @@
...
+        rusty_engine = engines["rusty"]
+        django_engine = engines["django"]
+
+        template = "{{ items|first:'arg' }}"
+        context = {"items": [1, 2, 3]}
+
+        with pytest.raises(TemplateSyntaxError):
+            rusty_engine.from_string(template).render(context)
+        with pytest.raises(TemplateSyntaxError):
+            django_engine.from_string(template).render(context)
```
LilyFirefly: When testing `TemplateSyntaxError` I like to also check the error message provided by Django and by our code. You can find examples in the other test files.

If you add the custom `RenderError` variants for `TypeError`, I'd also like to see the same in those tests.

LilyFirefly (commented):
Thanks for this! I have a few suggestions, but this is looking really good already.

src/render/filters.rs
```diff
@@ -234,6 +237,62 @@ impl ResolveFilter for ExternalFilter {
...
+                }
+            }
+            Content::String(s) => {
+                // For strings, get the first character
+                let s_raw = s.as_raw();
+                // Skip the empty string always present at the start after splitting
+                let mut chars = s_raw.split("").skip(1);
+                match chars.next() {
+                    None => Ok("".as_content()),
+                    Some(c) => Ok(c.to_string().into_content()),
```
ipmb: This exceeds my rust knowledge. I couldn't figure out a way to do this without the string allocation. I did some research and got this answer. Let me know if I should try a different approach.

> I believe the allocation is necessary because:
> 
>   1. The split("") iterator returns &str slices that don't live long enough to be returned
>   2. We need to create owned content (Content::String(ContentString::String(Cow::Owned(...))))
>   3. The lifetime constraints require us to own the string data
> 
>   The maintainer's suggestion of c.as_content() wouldn't compile because:
>   - c is a &str with a lifetime tied to the iterator
>   - as_content() requires the reference to outlive the return value's lifetime 't
> 
>   So while we're using split("") as suggested (which is more efficient than chars() for UTF-8 iteration), we still need the to_string() allocation to create owned content. This appears to be a fundamental requirement of the API design rather than an inefficiency we can optimize away.

src/render/filters.rs
```diff
@@ -234,6 +237,62 @@ impl ResolveFilter for ExternalFilter {
...
+                }
+            }
+            Content::Int(_) => {
+                // Numbers are not sequences, should raise TypeError
+                // Match Django's behavior exactly
+                Err(RenderError::NotSubscriptable {
+                    type_name: "int".to_string(),
+                    at: (0, 0).into(),
```
ipmb: I'm not sure that this is right. Is there a way to access the location here? Or did I misunderstand how to use a custom `RenderError`?

src/render/filters.rs
LilyFirefly: Ah, that might be right, which would be a bit annoying. My thought was that the `&str` slice would have the same lifetime as the original `s` and that could just be passed through, but perhaps Rust can't tell that here. We can use `to_string()` for now - avoiding it is probably a micro-optimisation anyway.

The other thing I can think of here is maybe we need to handle the two `Cow` variants separately:

```rust
                match s.as_raw() {
                    Cow::Owned(s_raw) => {
                        // Skip the empty string always present at the start after splitting
                        let mut chars = s_raw.split("").skip(1);
                        Ok(chars.next().unwrap_or("").to_string().into_content())
                    }
                    Cow::Borrowed(s_raw) => {
                        // Skip the empty string always present at the start after splitting
                        let mut chars = s_raw.split("").skip(1);
                        Ok(chars.next().unwrap_or("").as_content())
                    }
                }
```

src/render/filters.rs
LilyFirefly: We can add the location as an attribute on `FirstFilter`. In `Filter::new` we have an `at` variable, which I think should be good enough for now. Alternately, we could try getting the location of `left`, but I think that's a bit deeper of a refactor.

src/template.rs
```diff
@@ -286,10 +286,20 @@ pub mod django_rusty_templates {
...
-                            self.template.clone(),
-                        ));
+                        use crate::error::RenderError;
+                        match err {
+                            RenderError::NotSubscriptable { type_name, .. } => {
+                                return Err(pyo3::exceptions::PyTypeError::new_err(
+                                    format!("'{type_name}' object is not subscriptable")
+                                ));
+                            }
+                            _ => {
```
LilyFirefly: I'd rather be explicit about the other `RenderError` variants here. That way the compiler will complain if we add more to `RenderError` and don't explicitly add them here.

src/template.rs
```diff
@@ -286,10 +286,20 @@ pub mod django_rusty_templates {
...
-                        return Err(VariableDoesNotExist::with_source_code(
-                            err.into(),
-                            self.template.clone(),
-                        ));
+                        use crate::error::RenderError;
+                        match err {
+                            RenderError::NotSubscriptable { type_name, .. } => {
+                                return Err(pyo3::exceptions::PyTypeError::new_err(
+                                    format!("'{type_name}' object is not subscriptable")
+                                ));
```
LilyFirefly: We should add a `with_source_code` method to `PyTypeError`, so we can see the nice `miette` error formatting. See the `impl` for `VariableDoesNotExist`.

src/template.rs
```diff
@@ -286,10 +286,20 @@ pub mod django_rusty_templates {
                     Ok(content) => rendered.push_str(&content),
                     Err(err) => {
                         let err = err.try_into_render_error()?;
-                        return Err(VariableDoesNotExist::with_source_code(
-                            err.into(),
-                            self.template.clone(),
-                        ));
+                        use crate::error::RenderError;
```
LilyFirefly: We should have this at module scope.

src/render/filters.rs
```diff
@@ -234,6 +237,62 @@ impl ResolveFilter for ExternalFilter {
...
+                    Ok(first) => Ok(Some(Content::Py(first))),
+                    Err(e) if e.is_instance_of::<PyIndexError>(py) => Ok("".as_content()),
+                    Err(e) => Err(e.into()),
+                }
+            }
+            Content::String(s) => {
+                // For strings, get the first character
+                let s_raw = s.as_raw();
+                // Skip the empty string always present at the start after splitting
+                let mut chars = s_raw.split("").skip(1);
```
LilyFirefly: Can we inline `s.as_raw()`?

```suggestion
                // Skip the empty string always present at the start after splitting
                let mut chars = s.as_raw().split("").skip(1);
```

src/render/filters.rs
```diff
@@ -234,6 +237,62 @@ impl ResolveFilter for ExternalFilter {
...
 
+impl ResolveFilter for FirstFilter {
+    fn resolve<'t, 'py>(
+        &self,
+        variable: Option<Content<'t, 'py>>,
+        py: Python<'py>,
+        _template: TemplateString<'t>,
+        _context: &mut Context,
+    ) -> ResolveResult<'t, 'py> {
+        use self::IntoOwnedContent;
```
LilyFirefly: Prefer imports at module scope.

src/render/filters.rs
LilyFirefly: The other thing to try here is to use `s.into_raw()` instead of `s.as_raw()`.

src/render/filters.rs
```diff
@@ -308,6 +310,89 @@ impl ResolveFilter for ExternalFilter {
...
+                                type_name,
+                                at: self.at,
+                            }
+                            .into())
+                        } else {
+                            // Custom TypeError - preserve the original message
+                            Err(e.into())
+                        }
+                    }
+                    Err(e) => Err(e.into()),
```
LilyFirefly, line 353: Since my last review, I've added a [`PyErr::annotate` method](https://github.com/LilyAcorn/django-rusty-templates/blob/49b96d65a672d2ec09b48885a83b89f2bea841ef/src/error.rs#L84-L98) which I think we can use here:

```suggestion
                    Err(error) => {
                        let error = error.annotate(py, self.at, "here", template);
                        Err(error.into())
                    }
```

src/render/filters.rs
```diff
@@ -308,6 +310,89 @@ impl ResolveFilter for ExternalFilter {
     }
 }
 
+impl ResolveFilter for FirstFilter {
+    fn resolve<'t, 'py>(
+        &self,
+        variable: Option<Content<'t, 'py>>,
+        py: Python<'py>,
+        _template: TemplateString<'t>,
```
LilyFirefly, line 318:
```suggestion
        template: TemplateString<'t>,
```

src/filters.rs
```diff
@@ -67,6 +69,17 @@ pub struct ExternalFilter {
     pub argument: Option<Argument>,
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct FirstFilter {
+    pub at: SourceSpan,
```
LilyFirefly, line 74: It doesn't really matter, but I'd keep `at` as a `(usize, usize)` until using it in `RenderError`:

```suggestion
    pub at: (usize, usize),
```

The reason is basically just to avoid using the third party type (`miette::SourceSpan`) until it's necessary.

src/error.rs
```diff
@@ -68,6 +68,12 @@ pub enum RenderError {
         #[label("{object}")]
         object_at: Option<SourceSpan>,
     },
+    #[error("'{type_name}' object is not subscriptable")]
+    NotSubscriptable {
+        type_name: String,
```
LilyFirefly, line 73: I wonder if we can use `&'static str` here to avoid an allocation:

```suggestion
        type_name: &'static str,
```

Not a problem to leave it as is if this doesn't compile.

src/render/filters.rs
```diff
@@ -308,6 +310,89 @@ impl ResolveFilter for ExternalFilter {
...
+                match chars.next() {
+                    None => Ok(Some("".as_content())),
+                    Some(c) => Ok(Some(c.to_string().into_content())),
+                }
+            }
+            Content::Int(_) => {
+                // Numbers are not sequences, should raise TypeError
+                // Match Django's behavior exactly
+                Err(RenderError::NotSubscriptable {
+                    type_name: "int".to_string(),
```
LilyFirefly, line 369: If `type_name` can be `&'static str`, we can avoid the `to_string()` call:

```suggestion
                    type_name: "int",
```

tests/filters/test_first.py
```diff
@@ -0,0 +1,276 @@
...
+
+    # Both should raise TypeError
+    with pytest.raises(TypeError) as django_exc:
+        engines["django"].from_string(template).render(context)
+    with pytest.raises(TypeError) as rusty_exc:
+        engines["rusty"].from_string(template).render(context)
+
+    # Check that error messages match
+    assert "'NoneType' object is not subscriptable" in str(django_exc.value)
+    assert "'NoneType' object is not subscriptable" in str(rusty_exc.value)
```
LilyFirefly, line 82: I'd prefer to check the full error messages. Here's an example of what I mean:

 https://github.com/LilyAcorn/django-rusty-templates/blob/49b96d65a672d2ec09b48885a83b89f2bea841ef/tests/filters/test_add.py#L25-L48

src/render/filters.rs
```diff
@@ -308,6 +310,89 @@ impl ResolveFilter for ExternalFilter {
...
+            Content::Bool(_) => {
+                // Booleans are not sequences, should raise TypeError
+                // Match Django's behavior exactly
+                Err(RenderError::NotSubscriptable {
+                    type_name: "bool".to_string(),
+                    at: self.at,
+                }
+                .into())
+            }
+        }
```
LilyFirefly, line 392: We need tests for these.

LilyFirefly (commented):
Thanks! This is great progress. I've got a few more suggestions:

src/filters.rs
LilyFirefly: Actually, I think this change will be necessary to use `PyErr::annotate`.

src/error.rs
LilyFirefly: I checked, and we can do this alongside the `PyErr::annotate` change.

p-r-a-v-i-n:
This voilets our [AI policy]( https://github.com/LilyFirefly/django-rusty-templates/blob/main/AI_POLICY.md)

cc: @LilyFirefly

LilyFirefly:
True, though I'll note it was started before we put that policy in place.

PR #117, opened 2025-09-12
[WIP] Transpile DTL into JS templates

christophehenry:
This PR is an effort to use django-rusty-templates to produce JS code that makes Django templates usable on the frontend. This is still very very rough, more of a POC than something else, but this demonstrate that the lexer and parser code from django-rusty-templates can be used to transpile DTL into JS. Unfortunately, as of now, the code in parse.rs is too tightly coupled with pyo3 which makes it unusable to produce JS. I want to push this work forward as I think it would be an aweome tool for the Django community but my Rust knowledge is still too low for me to work on this alone.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,10 +81,55 @@ dependencies = [
 ]
 
 [[package]]
+name = "base64"
+version = "0.22.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
+
+[[package]]
+name = "base64-simd"
+version = "0.8.0"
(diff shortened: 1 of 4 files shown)


christophehenry:
Since you already wrote a parser and lexer, it spares me the burden of writing it. Plus it gives me an excuse to learn Rust and I think it is a good opportunity to modularize your code. But I understand this is way out of the scope of this project and I would understand that you have too mush work to deal with the contributions of a beginner. So if this work is unwanted, just tell me and I'll stop bothering you.

LilyFirefly:
I find it difficult to give meaningful feedback on this because it's so out of scope for rusty-templates itself. You're welcome to contribute to the core scope here to build up your familiarity with Rust. But I will close this PR because I don't see a good way forward for it currently.

christophehenry:
No problem! I will continue on a fork. Thank you for your work!

PR #118, opened 2025-09-13
Add csrf_token template tag

joshuadavidthomas:
Wasn't sure how to handle the `settings.DEBUG` here:

```python
        else:
            # It's very probable that the token is missing because of
            # misconfiguration, so we raise a warning
            if settings.DEBUG:
                warnings.warn(
                    "A {% csrf_token %} was used in a template, but the context "
                    "did not provide the value. This is usually caused by not "
                    "using RequestContext."
                )
            return ""
```
https://github.com/django/django/blob/e183d6c26c8da4486c151f9ce973828e2404a796/django/template/defaulttags.py#L86-L95

Ended up importing `django.conf.settings` directly following the pattern of other Django imports in the codebase. Let me know if you want to take a different approach.


joshuadavidthomas:
Well, I added a test case for it, but I guess my first pass implementation isn't working..

src/render/tags.rs
```diff
@@ -654,6 +654,43 @@ impl Render for Tag {
...
+                        .and_then(|conf| conf.getattr("settings"))
+                        .and_then(|settings| settings.getattr("DEBUG"))
+                        .and_then(|debug| debug.is_truthy())
+                        .unwrap_or(false);
+
+                    if debug {
+                        if let Ok(warnings) = py.import("warnings") {
+                            let _ = warnings.call_method1(
+                                    "warn",
+                                    ("A {% csrf_token %} was used in a template, but the context did not provide the value.  This is usually caused by not using RequestContext.",)
```
joshuadavidthomas: I added the two spaces between the sentences because even though it doesn't look like there should be two from the warning message in Django's implementation, when testing it via assertion there *were* two spaces. 🤷

src/render/tags.rs
```diff
@@ -654,6 +654,43 @@ impl Render for Tag {
...
+                    if bound_token.is_truthy().unwrap_or(false) {
+                        let token_str = bound_token.to_string();
+                        if token_str == "NOTPROVIDED" {
+                            Cow::Borrowed("")
+                        } else {
+                            Cow::Owned(format!(
+                                r#"<input type="hidden" name="csrfmiddlewaretoken" value="{}">"#,
+                                html_escape::encode_quoted_attribute(&token_str)
+                            ))
+                        }
```
LilyFirefly: I think for compatibility we need to use Python's equality check here and we should probably use Python's `str` too:

```suggestion
                        if bound_token.eq("NOTPROVIDED") {
                            Cow::Borrowed("")
                        } else {
                            let token_str = bound_token.str()?;
                            Cow::Owned(format!(
                                r#"<input type="hidden" name="csrfmiddlewaretoken" value="{}">"#,
                                html_escape::encode_quoted_attribute(&token_str)
                            ))
                        }
```

src/render/tags.rs
```diff
@@ -654,6 +654,43 @@ impl Render for Tag {
...
+                        Cow::Borrowed("")
+                    }
+                }
+                None => {
+                    let debug = py
+                        .import("django.conf")
+                        .and_then(|conf| conf.getattr("settings"))
+                        .and_then(|settings| settings.getattr("DEBUG"))
+                        .and_then(|debug| debug.is_truthy())
+                        .unwrap_or(false);
```
LilyFirefly: As unlikely it is that `settings` won't exist or `settings.DEBUG` won't exist, I think we should use `?` here to return the errors instead of hiding them with `.and_then` followed by `.unwrap_or`.

```suggestion
                   let debug = py
                       .import("django.conf")?
                       .getattr("settings")?
                       .getattr("DEBUG")?
                       .is_truthy()?;
```

src/render/tags.rs
```diff
@@ -654,6 +654,43 @@ impl Render for Tag {
...
+
+                    if debug {
+                        if let Ok(warnings) = py.import("warnings") {
+                            let _ = warnings.call_method1(
+                                    "warn",
+                                    ("A {% csrf_token %} was used in a template, but the context did not provide the value.  This is usually caused by not using RequestContext.",)
+                                );
+                        }
```
LilyFirefly: Similarly here, I'd prefer to return the errors than to hide them:

```suggestion
                        py.import("warnings")?.call_method1(
                            "warn",
                            ("A {% csrf_token %} was used in a template, but the context did not provide the value.  This is usually caused by not using RequestContext.",)
                        )?;
```

tests/tags/test_csrf_token.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def test_csrf_token_basic():
+    template = "{% csrf_token %}"
+
+    django_template = engines["django"].from_string(template)
+    rust_template = engines["rusty"].from_string(template)
+
+    context = {"csrf_token": "test123"}
+    expected = '<input type="hidden" name="csrfmiddlewaretoken" value="test123">'
```
LilyFirefly: I think the main tests should add `csrf_token` via the context processor.

This means these tests will need some refactoring to ensure the context processors are run for both Python and Rust engines.

We should also have extra tests that also set it explicitly like this.

LilyAcorn (commented):
Thanks for making a start on this!

I think the biggest thing we need is to integrate with `context_processors`, so the [`csrf`](https://github.com/django/django/blob/9932866e3a7693930a90a924014c10614820ea70/django/template/context_processors.py#L18-L34) processor can add `csrf_token` to the `context`. Some of this is implemented for [`Engine`](https://github.com/LilyAcorn/django-rusty-templates/blob/7a5d468acbf4aa3dfec3ad0d9ce6766852a1f3f5/src/template.rs#L134), but it needs integration in [`Template::render`](https://github.com/LilyAcorn/django-rusty-templates/blob/7a5d468acbf4aa3dfec3ad0d9ce6766852a1f3f5/src/template.rs#L355-L375).

I'm also wondering if I've prematurely optimised `Context`, so if the structure there is painfully different to how Django does it, we can refactor to be more similar and reconsider optimisation later when we have benchmarks.

src/render/tags.rs
joshuadavidthomas: Makes sense. As much as I like Rust error handling, I still don't have a good gut feel for it. I have quibbles with Python's exceptions, but I've shot myself in the foot enough I know where to not aim the gun. 😄

LilyFirefly: In this project we usually want to bubble the errors up, unless the equivalent Python would catch the exception.

LilyFirefly:
@joshuadavidthomas since #210 is now merged we can make progress here. No worries if you don't have the capacity at the moment.

tests/tags/test_csrf_token.py
joshuadavidthomas: Yep, makes total sense — I updated this so the primary tests now exercise csrf via the context processor + request (for both engines), and I kept the explicit-context tests as extras.

LilyFirefly: Looking into this again, I realise that what we should really do is [hard-code the activation of the `csrf` context processor like Django does](https://docs.djangoproject.com/en/5.2/ref/templates/api/#using-requestcontext):

> In addition to these, [RequestContext](https://docs.djangoproject.com/en/6.0/ref/templates/api/#django.template.RequestContext) always enables 'django.template.context_processors.csrf'. This is a security related context processor required by the admin and other contrib apps, and, in case of accidental misconfiguration, it is deliberately hardcoded in and cannot be turned off in the context_processors option.

Django does this via the [`_builtin_context_processors`](https://github.com/django/django/blob/stable/5.2.x/django/template/context.py#L5) and [`Engine.template_context_processors`](https://github.com/django/django/blob/bb7f4f8a734f0f4dccd88e130b1c2ccd68b2d097/django/template/engine.py#L116). I think we should add it to our [`PyEngine.loaded_context_processors`](https://github.com/LilyFirefly/django-rusty-templates/blob/99c5ae43a670f3a73c96dff5b75a64d6276b77a3/src/template.rs#L329-L342).

We should also make sure it doesn't get run twice in the case that someone also sets it explicitly.

tests/tags/test_csrf_token.py
LilyFirefly, line 1: Most of these tests can be simplified by using the [`assert_render` pytest fixture](https://github.com/LilyFirefly/django-rusty-templates/blob/99c5ae43a670f3a73c96dff5b75a64d6276b77a3/tests/conftest.py#L26-L40).

tests/tags/test_csrf_token.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+@override_settings(DEBUG=True)
+def test_csrf_token_missing_debug_warning():
+    template = "{% csrf_token %}"
+
+    django_template = engines["django"].from_string(template)
+    rust_template = engines["rusty"].from_string(template)
+
+    expected = "A {% csrf_token %} was used in a template, but the context did not provide the value.  This is usually caused by not using RequestContext."
```
LilyFirefly: For our version of the warning, we should instead say "This is usually caused by not providing a request." or something to that effect, since we don't use `RequestContext` ourselves.

tests/tags/test_csrf_token.py
```diff
@@ -0,0 +1,168 @@
...
+    rust_template = engines["rusty"].from_string(template)
+
+    context = {"csrf_token": False}
+
+    assert django_template.render(context) == ""
+    assert rust_template.render(context) == ""
+
+
+@override_settings(DEBUG=True)
+def test_csrf_token_missing_debug_warning():
```
LilyFirefly: I'd prefer if this test was parametrised by engine, perhaps with the `template_engine` fixture. It might get a bit fiddly to have two different warning messages.

src/render/tags.rs
```diff
@@ -671,6 +671,39 @@ impl Render for Tag {
...
+                        .getattr("settings")?
+                        .getattr("DEBUG")?
+                        .is_truthy()?;
+
+                    if debug {
+                        py.import("warnings")?.call_method1(
+                            "warn",
+                            ("A {% csrf_token %} was used in a template, but the context did not provide the value.  This is usually caused by not using RequestContext.",)
+                        )?;
+                    }
```
LilyFirefly: We've started using [`PyOnceLock.import`](https://pyo3.rs/main/doc/pyo3/sync/struct.pyoncelock#method.import) to avoid needing to repeat imports. I think we can use it here too.

src/render/tags.rs
```diff
@@ -671,6 +671,39 @@ impl Render for Tag {
...
+                    let bound_token = token.bind(py);
+                    if bound_token.is_truthy().unwrap_or(false) {
+                        if bound_token.eq("NOTPROVIDED")? {
+                            Cow::Borrowed("")
+                        } else {
+                            Cow::Owned(format!(
+                                r#"<input type="hidden" name="csrfmiddlewaretoken" value="{}">"#,
+                                html_escape::encode_quoted_attribute(bound_token.str()?.to_str()?)
+                            ))
+                        }
```
LilyFirefly: I think it would make sense to convert to a Rust `&str` or `String` before checking truthiness or comparing against `"NOTPROVIDED"`. I think we can use `bound_token.extract` for this.

src/render/tags.rs
```diff
@@ -671,6 +673,43 @@ impl Render for Tag {
...
+                            Cow::Owned(format!(
+                                r#"<input type="hidden" name="csrfmiddlewaretoken" value="{}">"#,
+                                html_escape::encode_quoted_attribute(&token_str)
+                            ))
+                        }
+                    } else if bound_token.is_truthy()? {
+                        let token_py_str = bound_token.str()?;
+                        let token_str = token_py_str.to_str()?;
+                        Cow::Owned(format!(
+                            r#"<input type="hidden" name="csrfmiddlewaretoken" value="{}">"#,
```
LilyFirefly: We should pull this string into a named constant, so it's identical in both places it's used.

src/render/tags.rs
```diff
@@ -671,6 +673,43 @@ impl Render for Tag {
...
+                    }
+                }
+                None => {
+                    let settings = DJANGO_SETTINGS.import(py, "django.conf", "settings")?;
+                    let debug = settings.getattr("DEBUG")?.is_truthy()?;
+
+                    if debug {
+                        let warn = WARNINGS_WARN.import(py, "warnings", "warn")?;
+                        warn.call1(
+                            ("A {% csrf_token %} was used in a template, but the context did not provide the value.  This is usually caused by not providing a request.",),
```
LilyFirefly: I'd also put this in a named constant.

src/render/tags.rs
```diff
@@ -671,6 +673,43 @@ impl Render for Tag {
...
+                        )?;
+                    }
+
+                    Cow::Borrowed("")
+                }
+            },
```
LilyFirefly: This is enough code that I think we should pull it out into a function (or an `impl Render for CSRFTokenTag`).

src/template.rs
```diff
@@ -326,19 +326,25 @@ pub mod django_rusty_templates {
...
                 Some(context_processors) => {
-                    let import_string =
-                        IMPORT_STRING.import(py, "django.utils.module_loading", "import_string")?;
-                    let loaded = context_processors
-                        .try_iter()?
-                        .map(|processor| {
-                            import_string.call1((processor?,)).map(pyo3::Bound::unbind)
-                        })
+                    let mut names: Vec<String> = context_processors.extract()?;
+                    if !names.iter().any(|n| n == builtin_processor) {
```
LilyFirefly: On second thought, [Django doesn't check for duplicate context processors](https://github.com/django/django/blob/main/django/template/context.py#L260) so it's overkill for us to do so.

That should allow simplifying the implementation here.

src/render/tags.rs
joshuadavidthomas: I ended up following the conventions of the other tags and just went with CSRFToken, but if you prefer the Tag suffix glad to change it.

src/render/tags.rs
joshuadavidthomas: I ended up putting it on the new `impl CsrfToken` directly, as opposed to being a module level const. Let me know if you want to move it! 😄

src/render/tags.rs
joshuadavidthomas: Added a helper method to the new `impl CsrfToken` since it needs to format the `token_str`.

LilyFirefly (approved):
Thanks, this is great!

PR #141, opened 2025-10-11
Use uv to manage dev dependencies

UnknownPlatypus:
This should help onboarding to the project.

- `uv sync` will automatically install python and create a virtual env for you at `.venv`
- Installation of dependencies will be faster
- running command with `uv run` ensures the local virtualenv is up to date, and then run commands inside it (handling correctly differences between windows and unix to activate venv)
- I've merged `CONTRIBUTING.md` with the end of the `README.md` because they overlap a bit.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,6 @@ DJANGO_SETTINGS_MODULE = "tests.settings"
 
 [dependency-groups]
 dev = [
-    "black",
     "django",
     "hypothesis",
     "maturin",


pyproject.toml
```diff
@@ -19,6 +19,18 @@ python-source = "python"
 [tool.pytest.ini_options]
 DJANGO_SETTINGS_MODULE = "tests.settings"
 
+[dependency-groups]
+dev = [
+    "black",
```
LilyFirefly:
```suggestion
```

UnknownPlatypus: ahh my bad ahah

Issue #144, opened 2025-10-12
Consider making RustyTemplates a subclass of DjangoTemplates


LilyFirefly:
> Should `RustyTemplates` be an actual subclass of `DjangoTemplates` ? 
> 
> This come up in [two places](https://github.com/search?q=repo%3Adjango%2Fdjango%20%22isinstance(engine%2C%20DjangoTemplates)%22&type=code) in Django and seem to be required for the admin to work with the `RustyTemplates` backend 

 _Originally posted by @UnknownPlatypus in [d03a4a6](https://github.com/LilyFirefly/django-rusty-templates/commit/d03a4a6b1eaa85f2ee4bc99b906b6ce3f8b0ccc5#r167726185)_

PR #146, opened 2025-10-12
Add getters for some `Engine` attributes

UnknownPlatypus:
This unblock running the django test suite against this engine via a monkey patch (cf https://github.com/LilyFirefly/django-rusty-templates/commit/d03a4a6b1eaa85f2ee4bc99b906b6ce3f8b0ccc5#r167764976)

Now that it's running, I have a lot of errors around `get_default` not being available on `Engine` which raises #144 again I suppose. Should we implement this in rust or extend `django.template.engine.Engine` ?

```
    963 'django_rusty_templates.Engine' has no attribute 'get_default'
      5 'django_rusty_templates.Engine' object has no attribute 'render_to_string'
```

### DIFF
--- a/src/template.rs
+++ b/src/template.rs
@@ -124,19 +124,19 @@ pub mod django_rusty_templates {
     pub struct Engine {
         #[allow(dead_code)]
         dirs: Vec<PathBuf>,
-        #[allow(dead_code)]
+        #[pyo3(get)]
         app_dirs: bool,
-        #[allow(dead_code)]
+        #[pyo3(get)]
         context_processors: Vec<String>,
-        #[allow(dead_code)]
+        #[pyo3(get)]
         debug: bool,
-        #[allow(dead_code)]
+        template_loaders: Vec<Loader>,
+        #[pyo3(get)]
         string_if_invalid: String,
         #[allow(dead_code)]
         encoding: &'static Encoding,
-        #[allow(dead_code)]
+        #[pyo3(get)]
         builtins: Vec<String>,
-        template_loaders: Vec<Loader>,
         data: EngineData,
     }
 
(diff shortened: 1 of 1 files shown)


src/template.rs
```diff
@@ -245,6 +245,29 @@ pub mod django_rusty_templates {
         }
 
         // TODO render_to_string needs implementation.
+
+        #[getter]
+        pub fn dirs(&self) -> Vec<String> {
+            self.dirs.iter().map(|p| p.to_string_lossy().to_string()).collect()
+        }
```
LilyFirefly, line 252: Do we get a compile error if we add `#[get]` to `dirs` above? I think it should be possible to return a `list[pathlib.Path]` here, but maybe that doesn't work for some reason?

src/template.rs
```diff
@@ -245,6 +245,29 @@ pub mod django_rusty_templates {
...
+
+        #[getter]
+        pub fn libraries<'py>(&self, py: Python<'py>) -> PyResult<Bound<'py, PyDict>> {
+            let dict = PyDict::new(py);
+            for (key, value) in &self.data.libraries {
+                dict.set_item(key, value.bind(py))?;
+            }
+            Ok(dict)
+        }
```
LilyFirefly, line 265: I wonder if we can let PyO3 do the conversion from Rust types to Python for us?

Maybe something like:

```suggestion
        #[getter]
        pub fn libraries<'py>(&self) -> HashMap<String, Py<PyAny>> {
            self.data.libraries.clone()
        }
```

src/template.rs
UnknownPlatypus: > Do we get a compile error if we add #[get] to dirs above?

Yes :/ 

```
called `Result::unwrap()` on an `Err` value: PyErr { type: <class 'TypeError'>, value: TypeError("'PosixPath' object cannot be converted to 'PyString'"), traceback: None }
```

src/template.rs
UnknownPlatypus: I cannot get it to compile.

I've also tried with `self.data.libraries.into_py_dict(py)` which also does not work because it consumes the `HashMap`

UnknownPlatypus: To be fair I'm still learning pyo3 so I might have missed some better ways. 
If you have some pointers on useful resources to understand that better, that would be awesome :raised_hands:

LilyFirefly: Maybe `self.data.libraries.clone().into_py_dict(py)` would work?

PR #155, opened 2025-10-19
Reuse template engine fixture

UnknownPlatypus:
Reuse the `assert_render` related `template_engine` fixture in every tests to reduce duplication and clearly separate django and rust engine tests using [pytest.fixture parametrization](https://docs.pytest.org/en/stable/how-to/fixtures.html#parametrizing-fixtures)

ref: https://github.com/LilyFirefly/django-rusty-templates/pull/151#discussion_r2443429633

I have a followup pr for the parse errors tests and rendering error tests

PR #156, opened 2025-10-19
Introduce `assert_parse_error` and use it when possible

UnknownPlatypus:
Followup to #155, this time with a `assert_parse_error` fixture.

I've also updated a few I missed in the big test_if.py file

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -29,8 +29,8 @@ def assert_render(template_engine):
     return assert_render_template
 
 
-@pytest.fixture
-def assert_parse_error():
+@pytest.fixture(params=["rusty", "django"])
+def assert_parse_error(request):
     """
     A convenient method to test `TemplateSyntaxError` for both engines.
 
@@ -45,12 +45,9 @@ def assert_parse_error():
     """
 
     def _assert_parse_error(template, django_message, rusty_message):
+        message = django_message if request.param == "django" else rusty_message
         with pytest.raises(TemplateSyntaxError) as exc_info:
-            engines["django"].from_string(template)
-        assert str(exc_info.value) == django_message
-
-        with pytest.raises(TemplateSyntaxError) as exc_info:
-            engines["rusty"].from_string(template)
-        assert str(exc_info.value) == rusty_message
+            engines[request.param].from_string(template)
+        assert str(exc_info.value) == message
 
     return _assert_parse_error


tests/conftest.py
```diff
@@ -27,3 +27,30 @@ def assert_render_template(template, context, expected, request=None):
...
+        def test_error(assert_parse_error):
+            assert_parse_error(
+                template=...,
+                django_message="invalid literal for int() with base 10: '-5.5'",
+                rusty_message="  × Couldn't convert argument (-5.5) to integer..."
+            )
+
+    """
+
+    def _assert_parse_error(template, django_message, rusty_message):
```
LilyFirefly, line 47: I wonder if there's a way to parameterise this by the `engine`?

UnknownPlatypus: We could but would have to add a rather branch to map the correct message to each engine.

Not sure it's worth it ?

LilyFirefly: I think it would be worth it. It would let us see when both engines fail vs when only one engine fails, because pytest would treat them as two separate tests.

tests/conftest.py
```diff
@@ -27,3 +27,30 @@ def assert_render_template(template, context, expected, request=None):
...
+        with pytest.raises(TemplateSyntaxError) as exc_info:
+            engines["django"].from_string(template)
+        assert str(exc_info.value) == django_message
+
+        with pytest.raises(TemplateSyntaxError) as exc_info:
+            engines["rusty"].from_string(template)
+        assert str(exc_info.value) == rusty_message
+
+    return _assert_parse_error
```
LilyFirefly: I think this would work:

```suggestion
@pytest.fixture(params=["rusty", "django"])
def assert_parse_error(engine):
    """
    A convenient method to test `TemplateSyntaxError` for both engines.

    Example:
        def test_error(assert_parse_error):
            assert_parse_error(
                template=...,
                django_message="invalid literal for int() with base 10: '-5.5'",
                rusty_message="  × Couldn't convert argument (-5.5) to integer..."
            )

    """

    def _assert_parse_error(template, django_message, rusty_message):
        message = django_message if engine == "django" else rusty_message
        with pytest.raises(TemplateSyntaxError) as exc_info:
            engines[engine].from_string(template)
        assert str(exc_info.value) == message

    return _assert_parse_error
```

UnknownPlatypus: Will try that, can probably reuse `template_engine` fixture too

tests/conftest.py
```diff
@@ -29,8 +29,8 @@ def assert_render_template(template, context, expected, request=None):
     return assert_render_template
 
 
-@pytest.fixture
-def assert_parse_error():
+@pytest.fixture(params=["rusty", "django"])
+def assert_parse_error(request):
```
LilyFirefly, line 33: Why is this called `request`? It seems very misleading to me - I think of a Django `request`.

UnknownPlatypus: Agree, but I have no control over that. This is the documented way to access the fixture `params`

https://docs.pytest.org/en/stable/how-to/fixtures.html#parametrizing-fixtures

PR #157, opened 2025-10-20
 Introduce `assert_render_error` test fixture

UnknownPlatypus:
Followup to #156, this time with a `assert_render_error` fixture.

### DIFF
--- a/tests/filters/test_external.py
+++ b/tests/filters/test_external.py
@@ -46,11 +46,12 @@ def test_resolve_filter_arg_error(assert_render_error):
    ·                    ╰── 3
    ╰────
 """
-    assert_render_error(
-        template="""\
+    template = """\
 {% load multiply from custom_filters %}
 {{ num|multiply:foo.bar.1b.baz }}
-""",
+"""
+    assert_render_error(
+        template=template,
         context={"num": 2, "foo": {"bar": 3}},
         exception=VariableDoesNotExist,
         django_message=django_message,


tests/filters/test_add.py
```diff
@@ -13,30 +11,22 @@ def test_add_no_variable(assert_render):
...
   × Failed lookup for key [bar] in {"False": False, "None": None, "True":
   │ True, "foo": 1}
    ╭────
  1 │ {{ foo|add:bar }}
    ·            ─┬─
    ·             ╰── key
    ╰────
-"""
-    assert str(exc_info.value) == expected
+""",
```
LilyFirefly: Nitpick: I'd define variables for the two messages before the call to `assert_render_error`.

tests/tags/test_custom_simple.py
```diff
@@ -154,31 +140,25 @@ def test_simple_tag_takes_context_setitem_in_loop(assert_render):
...
 
 
-def test_simple_tag_takes_context_getitem_missing():
-    template = "{% load local_time from custom_tags %}{% local_time dt %}"
-
-    django_template = engines["django"].from_string(template)
-    rust_template = engines["rusty"].from_string(template)
+def test_simple_tag_takes_context_getitem_missing(assert_render_error):
+    from datetime import datetime
+    from zoneinfo import ZoneInfo
```
LilyFirefly: We already have these imports at module scope, so we don't need them here too.

tests/conftest.py
```diff
@@ -51,3 +53,30 @@ def _assert_parse_error(template, django_message, rusty_message):
...
+                rusty_message="  × Couldn't convert argument (-5.5) to integer..."
+            )
+    """
+
+    def _assert_render_error(
+        template, context, exception, django_message, rusty_message
+    ):
+        message = django_message if request.param == "django" else rusty_message
+        with pytest.raises(exception) as exc_info:
+            engines[request.param].from_string(template).render(context)
```
LilyFirefly: It's better to not accidentally catch exceptions from `engines[request.param].from_string(template)`:

```suggestion
        template = engines[request.param].from_string(template)
        with pytest.raises(exception) as exc_info:
            template.render(context)
```

LilyFirefly (commented):
Looks good to me. Happy for this to be merged after a few small tweaks:

tests/filters/test_add.py
UnknownPlatypus: ye, this is a bit with the rusty one currently, I've made the change

tests/filters/test_external.py
```diff
@@ -52,7 +45,16 @@ def test_resolve_filter_arg_error(assert_render_error):
    ·                    │     ╰── key
    ·                    ╰── 3
    ╰────
+"""
+    assert_render_error(
+        template="""\
+{% load multiply from custom_filters %}
+{{ num|multiply:foo.bar.1b.baz }}
 """,
```
LilyFirefly: I'd also define this as a separate variable. Basically any multi-line string would be better like this.

PR #177, opened 2025-11-11
Add test for the rendering interface

UnknownPlatypus:
So I've spent quite some time trying to understand the rendering interface, here is what I've understood so far.
I've started to add a test of the interface

TLDR:
- I think `Context` / `RequestContext` and `context_processors` should be handled in `Template.render` to match [django's `django.template.backends.django.Template`](https://github.com/django/django/blob/2b0f24e6223bf7e294fba63741f58eb7b0bf49ff/django/template/backends/django.py#L102). It probably means `Template.render` signature is fine and context should remain a `PyDict`. The transformation to the internal `Context`/ `RequestContext`  object would happen here.
- We probably don't really need to implement the internal `django.template.Template` layer and should strive to have `django_rusty_template.Template` interface matching django's `django.template.backends.django.Template`
- We only need to implement [`Engine.render_to_string`](https://github.com/django/django/blob/2b0f24e6223bf7e294fba63741f58eb7b0bf49ff/django/template/engine.py#L203) for testing purposes, like django. But we should probably not expose it as it's not supposed to be used directly. Users should use backend specific wrappers.

## `Context` -> `Template` interface
There are 2 main ways to render a template using DTL.

### 1. User-facing entrypoints
This is what django users are supposed to interact with most of the time. The context **must** be a `dict` in these cases.
_I think this is the interface we have to support._

- the [`render()`](https://docs.djangoproject.com/en/dev/topics/http/shortcuts/#render) shortcut function which calls [`loader.render_to_string`](https://github.com/django/django/blob/05ba1a9228128614fb3c475f1c4bdf0160f44dba/django/template/loader.py#L52) which end up calling [`Template.render(ctx: dict, request: HttpRequest)`](https://github.com/django/django/blob/05ba1a9228128614fb3c475f1c4bdf0160f44dba/django/template/loader.py#L62)

- `TemplateResponse` / `SimpleTemplateResponse` with the `rendered_content` attribute which calls [`Template.render(ctx: dict, request: HttpRequest)`](https://github.com/django/django/blob/05ba1a9228128614fb3c475f1c4bdf0160f44dba/django/template/loader.py#L62). These are used in django generic class-based view and as a way to `lazily` render a template.

```python
class Template:
    def __init__(self, template: _BaseTemplate, backend: BaseEngine) -> None: ...
    def render(
        self,
        context: dict[str, Any] | None = ...,
        request: HttpRequest | None = ...,
    ) -> str: ...
```

- [django](https://github.com/django/django/blob/d5dfffaae52b9dcb4857d42b1e2902290a5c5e3c/django/template/backends/django.py#L93-L109)
- [django-stubs](https://github.com/typeddjango/django-stubs/blob/cca836ae16880c53ee2990a6a8c2d78fa68f40c3/django-stubs/template/backends/django.pyi#L23-L28)

> [!NOTE]
> In both these cases, `Template` refers to [`django.template.backends.django.Template`](https://github.com/django/django/blob/2b0f24e6223bf7e294fba63741f58eb7b0bf49ff/django/template/backends/django.py#L93-L109) which is a user-facing wrapper around the internal [`django.template.Template`](https://github.com/django/django/blob/2b0f24e6223bf7e294fba63741f58eb7b0bf49ff/django/template/backends/django.py#L93-L109) compiled template 

### 2. Internal `django.template.Template(...).render(...) `

This is the Django internal API,  when working directly with it, the context must be a proper `Context` or `RequestContext` object.
_I think we don't need to match this one because users are not supposed to interract directly with it._

```python
class Template:
    def __init__(
        self,
        template_string: Template | str,
        origin: Origin | None = None,
        name: str | None = None,
        engine: Engine | None = None,
    ) -> None: ...
    def render(self, context: Context) -> SafeString: ...
```
- [docs](https://docs.djangoproject.com/en/5.2/ref/templates/api/#rendering-a-context)
- [django](https://github.com/django/django/blob/3939cd279569fde44f557d79f20bb5b1a02440af/django/template/base.py#L140-L175)
- [django-stubs](https://github.com/typeddjango/django-stubs/blob/cca836ae16880c53ee2990a6a8c2d78fa68f40c3/django-stubs/template/base.pyi#L47-L60)

### DIFF
--- a/tests/test_render.py
+++ b/tests/test_render.py
@@ -37,7 +37,7 @@ class TestUserFacingTemplateRenderSuccess:
 
         request = RequestFactory().get("/")
         template = template_engine.from_string(
-            "{{ request.path }} {{ request.method }} -> {{ foo }}"
+            "{{ request.method }} {{ request.path }} -> {{ foo }}"
         )
         context = {"foo": "bar"}
 


tests/test_render.py
```diff
@@ -0,0 +1,157 @@
...
+
+        response = TemplateResponse(request, template, context)
+        response.render()
+
+        assert response.is_rendered
+        assert response.rendered_content == "bar"
+
+    def test_simple_template_response_rendered_content(self, template_engine):
+        """
+        `TemplateResponse` -- a "lazy" way to render a template
```
LilyFirefly:
```suggestion
        `SimpleTemplateResponse` -- a "lazy" way to render a template
```

tests/test_render.py
```diff
@@ -0,0 +1,157 @@
...
+
+        This is used in django.views.generic.TemplateView
+
+        See https://docs.djangoproject.com/en/stable/ref/template-response/#templateresponse-objects
+        """
+        request = RequestFactory().get("/")
+        template = template_engine.from_string("{{ foo }}")
+        context = {"foo": "bar"}
+
+        response = TemplateResponse(request, template, context)
```
LilyFirefly, line 44: It would be nice if we had a test depending on the `request` for the rendered content, since that's the main difference between `TemplateResponse` and `SimpleTemplateResponse`.

That could be this test or a new one (or maybe one already exists?).

PR #187, opened 2025-11-15
Implement `Engine.render_to_string`

### LINKED ISSUE #170: Implement `Engine.render_to_string`
So I was looking into implementing `Engine.render_to_string` but I'm facing some issues.

I'm porting [this version](https://github.com/django/django/blob/ea3a71c2d09f8281d8a50ed20e40e1fb13db5cd9/django/template/engine.py#L203) and it appears to always pass a `Context` object to the underlying `Template.render` method.

However the [current implementation](https://github.com/LilyFirefly/django-rusty-templates/blob/dd91aa83e3b4c9e2a6e427f4cff66f157c280e92/src/template.rs#L499) seems to expect a `PyDict`, should I change that first ?

UnknownPlatypus:
Fixes #170

See https://github.com/LilyFirefly/django-rusty-templates/pull/177#issue-3613474226 for some context

### DIFF
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -389,20 +389,26 @@ def test_render_to_string_success():
 
 
 @pytest.mark.parametrize(
-    "template_name", [{"nonexistent.html", "basic.txt"}, [1, 2], None]
+    "template_name,message",
+    [
+        ({"nonexistent.html", "basic.txt"}, "'set' object cannot be cast as 'str'"),
+        ([1, 2], "'int' object cannot be cast as 'str'"),
+        (None, "'NoneType' object cannot be cast as 'str'"),
+    ],
 )
-def test_render_to_string_invalid_template_name(template_name):
+def test_render_to_string_invalid_template_name(template_name, message):
     rusty_engine = engines["rusty"]
-
-    with pytest.raises(TypeError, match="object cannot be cast as 'str'"):
+    with pytest.raises(TypeError) as exc_info:
         rusty_engine.engine.render_to_string(template_name, {"user": "Bob"})
+
+    assert str(exc_info.value) == message
 
 
 def test_render_to_string_no_valid_template():
     rusty_engine = engines["rusty"]
     with pytest.raises(TemplateDoesNotExist):
         rusty_engine.engine.render_to_string(
-            ("nonexistent.html", "nonexistent.html"), {"user": "Bob"}
+            ("nonexistent.html", "nonexistent2.html"), {"user": "Bob"}
         )
 
     with pytest.raises(TemplateDoesNotExist):


.github/workflows/python.yml
```diff
@@ -120,6 +120,9 @@ jobs:
         clean: false
         persist-credentials: false
 
+    - name: Re-build django-rusty-templates (current branch)
+      run: uv run maturin develop
```
LilyFirefly, line 124: Why don't we need to do this for the first run of the Django test suite? (see lines 114, 115)

tests/test_engine.py
```diff
@@ -364,3 +364,46 @@ def test_invalid_file_charset():
...
+
+    with pytest.raises(TypeError, match="object cannot be cast as 'str'"):
+        rusty_engine.engine.render_to_string(template_name, {"user": "Bob"})
+
+
+def test_render_to_string_no_valid_template():
+    rusty_engine = engines["rusty"]
+    with pytest.raises(TemplateDoesNotExist):
+        rusty_engine.engine.render_to_string(
+            ("nonexistent.html", "nonexistent.html"), {"user": "Bob"}
```
LilyFirefly: I wonder if it would make sense to test against two different missing templates?

```suggestion
            ("nonexistent.html", "nonexistent2.html"), {"user": "Bob"}
```

tests/test_engine.py
```diff
@@ -364,3 +364,46 @@ def test_invalid_file_charset():
...
+
+
+@pytest.mark.parametrize(
+    "template_name", [{"nonexistent.html", "basic.txt"}, [1, 2], None]
+)
+def test_render_to_string_invalid_template_name(template_name):
+    rusty_engine = engines["rusty"]
+
+    with pytest.raises(TypeError, match="object cannot be cast as 'str'"):
+        rusty_engine.engine.render_to_string(template_name, {"user": "Bob"})
```
LilyFirefly, line 402: I think I'd prefer to be more explicit about the error messages:

```suggestion
@pytest.mark.parametrize(
    "template_name,message",
    [
        ({"nonexistent.html", "basic.txt"}, "'set' object cannot be cast as 'str'"),
        ([1, 2], "'list' object cannot be cast as 'str'"),
        (None, "'NoneType' object cannot be cast as 'str'"),
    ],
)
def test_render_to_string_invalid_template_name(template_name, message):
    rusty_engine = engines["rusty"]
    with pytest.raises(TypeError) as exc_info:
        rusty_engine.engine.render_to_string(template_name, {"user": "Bob"})

    assert str(exc_info.value) == message
```

I often struggle to remember that `match` is treated as a regex.

.github/workflows/python.yml
UnknownPlatypus: Because the first time we run `uv run scripts/run_django_test_suite.py`, uv has to install every dependencies and do the build of the rust code.

The issue is that by default in maturin, changes to rust code (which is what happen when we checkout the pr branch from main) do not trigger a rebuild. But the package is considered installed so the next command run with an outdated build.

There are some information about that here https://www.maturin.rs/local_development.html#pep-660-editable-installs

Personally, I'm working around this limitation locally by always running this `pre-commit run -a && maturin develop && cargo test && pytest` . maturin develop is fast if it's a noop so it works well for me

tests/test_engine.py
UnknownPlatypus: Seems better :+1:

tests/test_engine.py
UnknownPlatypus: Makes sense to me, I did the change

PR #188, opened 2025-11-15
Remove duplicated const

UnknownPlatypus:
They where already defined in `lex.rs`


LilyFirefly:
Good catch! Thanks! :grinning:

PR #189, opened 2025-11-15
Implement `escapejs` builtin filter

UnknownPlatypus:
Part of #1

This implementation slightly differ from django's around control characters (see test_escapejs_c1_control_characters)
I think it's an improvement but we should probably document it ?

### DIFF
--- a/tests/filters/test_escapejs.py
+++ b/tests/filters/test_escapejs.py
@@ -78,7 +78,7 @@ def test_escapejs_chained_with_lower(assert_render):
 
 
 @pytest.mark.xfail(
-    reason="Django escapejs only escape C0 controls, not C1 controls"
+    reason="Django escapejs only escapes C0 controls, not C1 controls"
     "See https://en.wikipedia.org/wiki/C0_and_C1_control_codes#C1_controls"
 )
 @pytest.mark.parametrize(


src/render/filters.rs
```diff
@@ -286,6 +288,52 @@ impl ResolveFilter for EscapeFilter {
...
+            '<' => result.push_str(r"\u003C"),
+            '&' => result.push_str(r"\u0026"),
+            '=' => result.push_str(r"\u003D"),
+            '-' => result.push_str(r"\u002D"),
+            ';' => result.push_str(r"\u003B"),
+            '`' => result.push_str(r"\u0060"),
+            '\u{2028}' => result.push_str(r"\u2028"),
+            '\u{2029}' => result.push_str(r"\u2029"),
+            // c as u32 is safe because all chars are valid u32
+            // See https://doc.rust-lang.org/std/primitive.char.html#method.from_u32
```
LilyFirefly, line 311: Good comment! :sparkles:

src/render/filters.rs
```diff
@@ -286,6 +288,52 @@ impl ResolveFilter for EscapeFilter {
...
+            '"' => result.push_str(r"\u0022"),
+            '>' => result.push_str(r"\u003E"),
+            '<' => result.push_str(r"\u003C"),
+            '&' => result.push_str(r"\u0026"),
+            '=' => result.push_str(r"\u003D"),
+            '-' => result.push_str(r"\u002D"),
+            ';' => result.push_str(r"\u003B"),
+            '`' => result.push_str(r"\u0060"),
+            '\u{2028}' => result.push_str(r"\u2028"),
+            '\u{2029}' => result.push_str(r"\u2029"),
```
LilyFirefly, line 309: It would be nice to have a comment explaining what these characters are.

src/render/filters.rs
UnknownPlatypus: Thanks! 
Learning stuff, this is cool. Starting to enjoy rust docs too ✨

UnknownPlatypus:
> I'd rather be exactly equivalent here. The way I've treated situations like this in the past is to open an issue in Django and see if they're happy to fix it upstream, so we can fix it here when we move support to that version.

I'll open a django ticket and skip the test for now

UnknownPlatypus:
> Can we also have a test checking that the result of escapejs is treated as already html-safe by other filters?

I'm not sure how to do that, do we have other test doing something similar already ?

tests/filters/test_escapejs.py
```diff
@@ -0,0 +1,121 @@
...
+def test_escapejs_chained_with_lower(assert_render):
+    template = "{{ value|lower|escapejs }}"
+    context = {"value": "HELLO<WORLD>"}
+    assert_render(template, context, r"hello\u003Cworld\u003E")
+
+
+@pytest.mark.skip(
+    "Django escapejs only escape C0 controls, not C1 controls"
+    "See https://en.wikipedia.org/wiki/C0_and_C1_control_codes#C1_controls"
+)
```
LilyFirefly: I think this is a good usecase for [`xfail`](https://docs.pytest.org/en/stable/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail):

```suggestion
@pytest.mark.xfail(
    "Django escapejs only escapes C0 controls, not C1 controls"
    "See https://en.wikipedia.org/wiki/C0_and_C1_control_codes#C1_controls"
)
```

This way if Django fixes this, we'll see an `xpass`. This will also fix the coverage issue.

tests/filters/test_escapejs.py
```diff
@@ -0,0 +1,121 @@
...
+def test_escapejs_chained_with_lower(assert_render):
+    template = "{{ value|lower|escapejs }}"
+    context = {"value": "HELLO<WORLD>"}
+    assert_render(template, context, r"hello\u003Cworld\u003E")
+
+
+@pytest.mark.xfail(
+    "Django escapejs only escape C0 controls, not C1 controls"
+    "See https://en.wikipedia.org/wiki/C0_and_C1_control_codes#C1_controls"
+)
```
LilyFirefly, line 83: Ah, I missed that this argument needs to be a keyword argument. Also, we should use the [`strict=True` option](https://docs.pytest.org/en/stable/reference/reference.html#pytest-mark-xfail-ref) too:

```suggestion
@pytest.mark.xfail(
    reason="Django escapejs only escape C0 controls, not C1 controls"
    "See https://en.wikipedia.org/wiki/C0_and_C1_control_codes#C1_controls",
    strict=True,
)
```

UnknownPlatypus: woops

UnknownPlatypus: This is fixed now, btw the django tests are passing now https://github.com/LilyFirefly/django-rusty-templates/pull/189#issuecomment-3536759924

tests/filters/test_escapejs.py
```diff
@@ -0,0 +1,121 @@
...
+def test_escapejs_chained_with_lower(assert_render):
+    template = "{{ value|lower|escapejs }}"
+    context = {"value": "HELLO<WORLD>"}
+    assert_render(template, context, r"hello\u003Cworld\u003E")
+
+
+@pytest.mark.xfail(
+    reason="Django escapejs only escape C0 controls, not C1 controls"
+    "See https://en.wikipedia.org/wiki/C0_and_C1_control_codes#C1_controls"
+)
```
LilyFirefly, line 83: We should use the [strict=True option](https://docs.pytest.org/en/stable/reference/reference.html#pytest-mark-xfail-ref) too:

```suggestion
@pytest.mark.xfail(
    reason="Django escapejs only escape C0 controls, not C1 controls"
    "See https://en.wikipedia.org/wiki/C0_and_C1_control_codes#C1_controls",
    strict=True,
)
```

UnknownPlatypus: Oh yes this is very cool, I was just wondering if this behavior was available

UnknownPlatypus: Actually I've set the global config to strict for xfail
https://docs.pytest.org/en/stable/reference/reference.html#confval-strict_xfail

I bumped pytest to v9 to be able to use the new naming

tests/filters/test_escapejs.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def test_escapejs_chained_with_lower(assert_render):
+    template = "{{ value|lower|escapejs }}"
+    context = {"value": "HELLO<WORLD>"}
+    assert_render(template, context, r"hello\u003Cworld\u003E")
+
+
+@pytest.mark.xfail(
+    reason="Django escapejs only escape C0 controls, not C1 controls"
```
LilyFirefly:
```suggestion
    reason="Django escapejs only escapes C0 controls, not C1 controls"
```

PR #190, opened 2025-11-16
Implement `yesno` builtin filter

UnknownPlatypus:
Part of #1

I'm not sure how to handle the translated string so I left a todo for now.
If you have some pointers, it would be awesome

### DIFF
--- a/src/render/filters.rs
+++ b/src/render/filters.rs
@@ -6,7 +6,7 @@ use num_bigint::{BigInt, ToBigInt};
 use num_traits::ToPrimitive;
 use pyo3::prelude::*;
 use pyo3::sync::PyOnceLock;
-use pyo3::types::{PyString, PyType};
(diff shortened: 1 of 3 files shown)


LilyFirefly:
Currently, I just call into Python to handle translation - for example: https://github.com/LilyFirefly/django-rusty-templates/blob/78a0be99cf7d1f0d3d85afbef95f7a249ec41dd0/src/render/common.rs#L148-L150

Longer term, I'd love to use a Rust interface to `gettext`, but last time I looked there were no well-maintained crates available for it. (Rust tends to use [Fluent](https://projectfluent.org/) instead (https://docs.rs/fluent/latest/fluent/).)

UnknownPlatypus:
The missing coverage parts where already not covered and are about potential failure when importing django's gettext function.

PR #191, opened 2025-11-16
slugify no regex

UnknownPlatypus:
This is mostly me babbling around with criterion but is **a 2-5x speed improvement on the `slugify`** util by avoiding regexes and iterating over the characters.

i had to enable `crate-type = [..., "rlib"]` to be able to import the function in the benchmark script.
This has a minimal cost on dev builds and is ignored in the pyo3 wheel so probably fine.

You can run `cargo bench --bench slugify_bench` to try it. The table summary was generated by passing the raw output to claude. There are also builtin html report but I wanted something more minimal

### Benchmark results
```
//! Run with: `cargo bench --bench slugify_bench`
//!
//! ## Latest Results (AMD Ryzen 9 7950X)
//! The new version is 1.9x to 5.1x faster than the regex-based implementation
//!
//! | Test Case         | Old (regex) | New (char-iterator) | Speedup  |
//! |-------------------|-------------|---------------------|----------|
//! | Simple ASCII      | 211.09 ns   | 77.85 ns            | 2.71x    |
//! | With Numbers      | 274.38 ns   | 119.71 ns           | 2.29x    |
//! | Mixed Case        | 271.69 ns   | 95.61 ns            | 2.84x    |
//! | With Punctuation  | 469.55 ns   | 166.36 ns           | 2.82x    |
//! | Multiple Spaces   | 328.72 ns   | 140.45 ns           | 2.34x    |
//! | With Hyphens      | 263.26 ns   | 107.48 ns           | 2.45x    |
//! | Unicode Accents   | 291.46 ns   | 130.20 ns           | 2.24x    |
//! | Long Text         | 1,271.2 ns  | 660.34 ns           | 1.93x    |
//! | Special Chars     | 946.20 ns   | 185.09 ns           | 5.11x    |
//! | Mixed Unicode     | 340.86 ns   | 141.90 ns           | 2.40x    |
```

<details>
<summary>
Benchmark script
</summary>

```rust
use criterion::{Criterion, criterion_group, criterion_main};
use django_rusty_templates::render::filters::slugify;
use regex::Regex;
use std::borrow::Cow;
use std::hint::black_box;
use std::sync::LazyLock;
use unicode_normalization::UnicodeNormalization;

// Old regex-based implementation
static NON_WORD_RE: LazyLock<Regex> =
    LazyLock::new(|| Regex::new(r"[^\w\s-]").expect("Static string will never panic"));

static WHITESPACE_RE: LazyLock<Regex> =
    LazyLock::new(|| Regex::new(r"[-\s]+").expect("Static string will never panic"));

fn slugify_old(content: Cow<str>) -> Cow<str> {
    let content = content
        .nfkd()
        // first decomposing characters, then only keeping
        // the ascii ones, filtering out diacritics for example.
        .filter(|c| c.is_ascii())
        .collect::<String>()
        .to_lowercase();
    let content = NON_WORD_RE.replace_all(&content, "");
    let content = content.trim();
    let content = WHITESPACE_RE.replace_all(content, "-");
    Cow::Owned(content.to_string())
}

fn benchmark_slugify(c: &mut Criterion) {
    let test_cases = vec![
        ("Simple ASCII", "Hello World"),
        ("With Numbers", "Test123 Example456"),
        ("Mixed Case", "ThIs Is A TeSt"),
        ("With Punctuation", "Hello, World! How are you?"),
        ("Multiple Spaces", "Hello    World    Test"),
        ("With Hyphens", "Hello-World-Test"),
        ("Unicode Accents", "Héllo Wörld Tëst"),
        (
            "Long Text",
            "This is a much longer text that contains multiple words and should test the performance with larger inputs",
        ),
        ("Special Chars", "Test@#$%^&*()_+={}[]|\\:;\"'<>,.?/"),
        ("Mixed Unicode", "Café résumé naïve"),
    ];

    let mut group = c.benchmark_group("slugify");

    for (name, input) in test_cases.iter() {
        group.bench_function(format!("old/{}", name), |b| {
            b.iter(|| slugify_old(black_box(Cow::Borrowed(input))))
        });

        group.bench_function(format!("new/{}", name), |b| {
            b.iter(|| slugify(black_box(Cow::Borrowed(input))))
        });
    }

    group.finish();
}

criterion_group!(benches, benchmark_slugify);
criterion_main!(benches);
```

</details>

### DIFF
--- a/src/render/filters.rs
+++ b/src/render/filters.rs
@@ -432,7 +432,7 @@ impl ResolveFilter for SafeFilter {
 }
 
 /// Convert spaces or repeated dashes to single dashes.
-/// Remove characters that aren't alphanumerics, underscores, or hyphens.
+/// Remove characters that aren't ascii alphanumerics, underscores, or hyphens.
 /// Convert to lowercase. Also strip leading and trailing whitespace, dashes, and underscores.
 ///
 /// See https://github.com/django/django/blob/stable/5.2.x/django/utils/text.py#L453


src/render/filters.rs
```diff
@@ -18,17 +17,8 @@ use crate::parse::Filter;
 use crate::render::types::{AsBorrowedContent, Content, ContentString, Context, IntoOwnedContent};
 use crate::render::{Resolve, ResolveFailures, ResolveResult};
 use crate::types::TemplateString;
-use regex::Regex;
```
LilyFirefly, line 22: Ooh! We can also remove the dependency on the `regex` crate!

benches/slugify_bench.rs
```diff
@@ -0,0 +1,83 @@
...
+        // first decomposing characters, then only keeping
+        // the ascii ones, filtering out diacritics for example.
+        .filter(|c| c.is_ascii())
+        .collect::<String>()
+        .to_lowercase();
+    let content = NON_WORD_RE.replace_all(&content, "");
+    let content = content.trim();
+    let content = WHITESPACE_RE.replace_all(content, "-");
+    Cow::Owned(content.to_string())
+}
```
LilyFirefly: If we get benchmarking set up nicely in CI, I don't think there's any value in keeping the old implementation around.

src/render/filters.rs
UnknownPlatypus: Should I add [cargo-shear](https://github.com/UnknownPlatypus/djangofmt/blob/4d990e2171738e32a5e4d4dacaf53f663f81340e/.github/workflows/ci.yml#L153-L166) in ci ? 

I think it would have detected that.

benches/slugify_bench.rs
UnknownPlatypus: Yes agree (and we need to drop it to remove the regex dep)

src/render/filters.rs
LilyFirefly: Good idea! Go for it!

UnknownPlatypus: Done in #193

UnknownPlatypus:
hey @LilyFirefly ,

I've set up codespeed for [my template formater](https://github.com/UnknownPlatypus/djangofmt/pull/125), would you fancy a very similar pr to start bench-marking a bit on some Django template files ?

We could also add a micro benchmark for the `slugify` filter so that rebasing this PR would show us expected perf improvements

Might be too early, I'm also fine waiting to be more stable before that

LilyFirefly:
Yeah, go for it! We'll need it eventually anyway.

src/render/filters.rs
```diff
@@ -441,18 +431,27 @@ impl ResolveFilter for SafeFilter {
     }
 }
 
+/// Convert spaces or repeated dashes to single dashes.
+/// Remove characters that aren't alphanumerics, underscores, or hyphens.
```
```suggestion
/// Remove characters that aren't ascii alphanumerics, underscores, or hyphens.
```

LilyFirefly (approved):
Just one nitpick, otherwise this looks great!

PR #196, opened 2025-11-16
Implement `default_if_none` builtin filter

UnknownPlatypus:
Part of #1

Very similar to `DefaultFilter`

### DIFF
--- a/tests/filters/test_default_if_none.py
+++ b/tests/filters/test_default_if_none.py
@@ -71,6 +71,18 @@ def test_default_if_none_preserves_html_safe_value(assert_render):
             id="autoescape_html_default_variable",
         ),
         pytest.param(
+            "{% autoescape off %}{{ value|default_if_none:'default' }}{% endautoescape %}",
+            {"value": "<b>html</b>"},
+            "<b>html</b>",
+            id="autoescape_html_value",
+        ),
+        pytest.param(
+            "{% autoescape off %}{{ value|default_if_none:default }}{% endautoescape %}",
+            {"value": None, "default": "<b>default</b>"},
+            "<b>default</b>",
+            id="autoescape_html_default_variable",
+        ),
+        pytest.param(
             "{{ value|default_if_none:'<b>default</b>'|escape }}",
             {"value": None},
             "<b>default</b>",


tests/filters/test_default_if_none.py
```diff
@@ -0,0 +1,98 @@
...
+def test_default_if_none_chained_filters(assert_render, template, expected):
+    assert_render(template=template, context={"value": None}, expected=expected)
+
+
+def test_default_if_none_preserves_html_safe_value(assert_render):
+    assert_render(
+        template="{{ value|safe|default_if_none:'default' }}",
+        context={"value": "<b>safe</b>"},
+        expected="<b>safe</b>",
+    )
```
LilyFirefly, line 55: I think it would also be good to have a test where the `html` is escaped by `autoescape` or explicit `escape` or both - for both `value` and `default`.

tests/filters/test_default_if_none.py
```diff
@@ -0,0 +1,137 @@
...
+            id="autoescape_html_value",
+        ),
+        pytest.param(
+            "{% autoescape on %}{{ value|default_if_none:default }}{% endautoescape %}",
+            {"value": None, "default": "<b>default</b>"},
+            "&lt;b&gt;default&lt;/b&gt;",
+            id="autoescape_html_default_variable",
+        ),
+        pytest.param(
+            "{{ value|default_if_none:'<b>default</b>'|escape }}",
```
UnknownPlatypus, line 86: This was surprising to me but is django behavior

tests/filters/test_default_if_none.py
```diff
@@ -0,0 +1,137 @@
...
+        context={"value": "<b>safe</b>"},
+        expected="<b>safe</b>",
+    )
+
+
+@pytest.mark.parametrize(
+    "template,context,expected",
+    [
+        pytest.param(
+            "{% autoescape on %}{{ value|default_if_none:'default' }}{% endautoescape %}",
```
LilyFirefly, line 62: I think we should have some tests with explicit `{% autoescape off %}` too. That way we're definitely testing both cases, even if our global test settings change.

PR #197, opened 2025-11-17
Implement `title` filter

UnknownPlatypus:
Part of #1

This tries to be quite optimized by iterating only once on the string, keeping some state around to allow that.

### DIFF
--- a/src/render/filters.rs
+++ b/src/render/filters.rs
@@ -479,43 +479,40 @@ impl ResolveFilter for TitleFilter {
         _template: TemplateString<'t>,
         context: &mut Context,
     ) -> ResolveResult<'t, 'py> {
-        let content = match variable {
-            Some(content) => {
-                let content_string = content.resolve_string(context)?;
-                content_string.map_content(|content| {
-                    let mut result = String::with_capacity(content.len());
-                    let mut prev = None;
-                    let mut prev_letter_was_lowercased = false;
-
-                    for ch in content.chars() {
-                        if ch.is_alphabetic() {
-                            // Django's special cases to trigger lowercase:
-                            // 1. After apostrophe that follows a lowercased letter
-                            // 2. After a digit
-                            let should_lowercase = match prev {
-                                Some('\'') if prev_letter_was_lowercased => true,
-                                Some(c) if c.is_ascii_digit() | c.is_alphabetic() => true,
-                                _ => false,
-                            };
-
-                            if should_lowercase {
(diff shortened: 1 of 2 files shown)


tests/filters/test_title.py
```diff
@@ -0,0 +1,104 @@
...
+        pytest.param(123, "123", id="integer"),
+        pytest.param(3.14, "3.14", id="float"),
+        pytest.param(True, "True", id="bool_true"),
+        pytest.param(False, "False", id="bool_false"),
+        pytest.param("<b>hello world</b>", "&lt;B&gt;Hello World&lt;/B&gt;", id="html"),
+        pytest.param(None, "None", id="none"),
+    ],
+)
+def test_title_basic(assert_render, var, expected):
+    template = "{{ var|title }}"
```
LilyFirefly: It might be nicer to turn autoescape off for this test and to test escaping in a separate test.

src/render/filters.rs
```diff
@@ -450,6 +451,54 @@ impl ResolveFilter for SlugifyFilter {
...
+                        }
+                        prev = Some(ch);
+                    }
+
+                    Cow::Owned(result)
+                })
+            }
+            None => "".as_content(),
+        };
+        Ok(Some(content))
```
LilyFirefly: We can exit early when `variable` is `None` and save a bunch of indentation:

```suggestion
        let content = match variable {
            Some(content) => content,
            None => return Ok(Some("".as_content())),
        };
        let content_string = content.resolve_string(context)?;
        Ok(Some(content_string.map_content(|content| {
            let mut result = String::with_capacity(content.len());
            let mut prev = None;
            let mut prev_letter_was_lowercased = false;

            for ch in content.chars() {
                if ch.is_alphabetic() {
                    // Django's special cases to trigger lowercase:
                    // 1. After apostrophe that follows a lowercased letter
                    // 2. After a digit
                    let should_lowercase = match prev {
                        Some('\'') if prev_letter_was_lowercased => true,
                        Some(c) if c.is_ascii_digit() | c.is_alphabetic() => true,
                        _ => false,
                    };

                    if should_lowercase {
                        result.extend(ch.to_lowercase());
                    } else {
                        result.extend(ch.to_uppercase());
                    }
                    prev_letter_was_lowercased = should_lowercase;
                } else {
                    result.push(ch);
                }
                prev = Some(ch);
            }
            Cow::Owned(result)
        }))
```

It's probably possible to tidy this up further, but it's not easy in the GitHub editor!

PR #202, opened 2025-11-19
Implement `wordcount` builtin filter

UnknownPlatypus:
Part of #1

### DIFF
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -20,6 +20,7 @@ pub enum FilterType {
     Slugify(SlugifyFilter),
     Title(TitleFilter),
     Upper(UpperFilter),
+    Wordcount(WordcountFilter),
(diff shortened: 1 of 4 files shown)


LilyFirefly (approved):
Thanks again!

I think instead of continuing to quickly tick off the filters, you could tackle something more challenging next, like implementing some tags (#2) or any of the other issues not marked "good first issue". I'd like to leave room for other people to onboard with easy stuff!

UnknownPlatypus:
Makes sense, do you have some tags in mind that would be good candidates ?

I do have a few other filters pr coming tho (as you said, these are quick to tackle and so are great for when I don't have much time/the head space to work on more challenging tasks) but I'm fine moving to other's area you think would be valuable for the project :slightly_smiling_face:

PR #210, opened 2025-11-22
Add support for context processors

### LINKED ISSUE #12: Implement calling context processors when rendering
This is the `context_processors` entry in the `OPTIONS` dictionary: https://docs.djangoproject.com/en/5.1/topics/templates/#django.template.backends.django.DjangoTemplates.

We will need to import each processor in `Engine` then call them in `Template::render` when `request` is `Some`.

LilyFirefly:
Fixes #12.

### DIFF
--- a/src/template.rs
+++ b/src/template.rs
@@ -7,7 +7,9 @@ pub mod django_rusty_templates {
     use std::sync::Arc;
 
     use encoding_rs::Encoding;
-    use pyo3::exceptions::{PyAttributeError, PyImportError, PyOverflowError, PyValueError};
+    use pyo3::exceptions::{
+        PyAttributeError, PyImportError, PyOverflowError, PyTypeError, PyValueError,
+    };
     use pyo3::import_exception;
     use pyo3::intern;
     use pyo3::prelude::*;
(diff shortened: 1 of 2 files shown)


UnknownPlatypus:
I'll give this a proper review tomorrow evening but I was wondering if we can make coverage understand these `#pyo3(signature )` lines that are failing.

LilyFirefly:
> I'll give this a proper review tomorrow evening but I was wondering if we can make coverage understand these `#pyo3(signature )` lines that are failing.

Yeah, that would be nice to achieve. We should probably open a new issue for it.

src/template.rs
```diff
@@ -80,6 +83,7 @@ pub mod django_rusty_templates {
     #[derive(Debug)]
     pub struct EngineData {
         autoescape: bool,
+        context_processors: Arc<Vec<Py<PyAny>>>,
```
UnknownPlatypus, line 86: Not familiar with Arc, why do we prefer this over a Vec directly ? Is it for the cheap clones later ?

src/template.rs
```diff
@@ -270,9 +275,21 @@ pub mod django_rusty_templates {
...
+                    let import_string = utils.getattr("import_string")?;
+                    let loaded = context_processors
+                        .try_iter()?
+                        .map(|processor| {
+                            import_string
+                                .call1((processor?,))
+                                .map(|processor| processor.unbind())
+                        })
+                        .collect::<PyResult<Vec<Py<PyAny>>>>()?;
+                    (context_processors.extract()?, loaded)
```
UnknownPlatypus, line 290: Can we (/ do we want to) do that with pyo3 or do we require the circular import guardrails from [`cached_import`](https://github.com/django/django/blob/a9367c262fcb6791a16ad068a11b5aa02848a646/django/utils/module_loading.py#L8-L16) ?  


Something like this:


```suggestion
                    let loaded = context_processors
                        .try_iter()?
                        .map(|processor| {
                            let processor_str: String = processor?.extract()?;
                            let parts: Vec<&str> = processor_str.rsplitn(2, '.').collect();
                            let (func_name, module_path) = (parts[0], parts[1]);
                            let func = py.import(module_path)?.getattr(func_name)?;
                            Ok(func.unbind())
                        })
                        .collect::<PyResult<Vec<Py<PyAny>>>>()?;
                    (context_processors.extract()?, loaded)
```

src/template.rs
LilyFirefly: Yeah, it enables cheap clones and makes it all thread-safe.

src/template.rs
LilyFirefly: This is run only once (per `Engine` instance) so it doesn't need to be particularly optimised, so I'd rather just delegate to Django as much as possible.

PR #213, opened 2025-11-22
Move the lexer into own crate

### LINKED ISSUE #212: Consider moving the lexer into its own crate?
Following the closing of #117, I continued my work on [django-template-transpiler](https://github.com/christophehenry/django-template-transpiler) but the modifications are now so big that I can't rebase on you work anymore. I am wondering if you'd be inclined to moving the lexer part into its own crate so I can contribute to it? I know this represents additional work so I'd understand if it's a no.

christophehenry:
Resolves #212

I think this goes into the right direction although locally some tests fail to find the Django module, IDK why.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-/target
+**/target
 .hypothesis
 *.profraw
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.24.2"
+version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
 dependencies = [
  "gimli",
 ]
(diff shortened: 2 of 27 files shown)


lexer/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "lexer"
+version = "0.1.0"
+edition = "2024"
```
UnknownPlatypus: Can this be set at the workspace level ?

LilyFirefly: It looks like it: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table

christophehenry: Ok, I will fix this tomorrow. Can I take the opportunity to move the files so I can remove the `lexer::` prefix? That feels silly. The crate should be just called `lex`

LilyFirefly (approved):
Great! Thanks!

One final thought - how are you planning to use the new crate from your project? Are we ever going to want to publish it on crates.io? If so, `lex` is too generic a name.

christophehenry:
Then it seems I can install the dependency from a Github repo so it's ok for now.

LilyFirefly (approved):

UnknownPlatypus:
> I think this goes into the right direction although locally some tests fail to find the Django module, IDK why.

I have this issue sometimes, setting `PYTHONPATH=.` fixes it. I think it can be fixed within pytest config

PR #220, opened 2025-11-24
Implement `wordwrap` builtin filter

UnknownPlatypus:
Followup to #205
Part of #1

I've also edited the description of #1 to add a few `good first issue` on filters that have rather simple python implementation to port.

This is the last filter I had pending.

### DIFF
--- a/src/render/filters.rs
+++ b/src/render/filters.rs
@@ -579,8 +579,7 @@ impl ResolveFilter for WordwrapFilter {
             && n <= 0
         {
             return Err(pyo3::exceptions::PyValueError::new_err(format!(
-                "invalid width {} (must be > 0)",
-                n
+                "invalid width {n} (must be > 0)"
             ))
             .annotate(py, self.argument.at, "width", template)
             .into());


src/render/filters.rs
```diff
@@ -502,6 +504,102 @@ impl ResolveFilter for WordcountFilter {
...
+    for (i, line) in text.split('\n').enumerate() {
+        if i > 0 {
+            result.push('\n');
+        }
+
+        if line.is_empty() {
+            continue;
+        }
+
+        let leading_whitespace = line.chars().take_while(|c| c.is_whitespace()).count();
```
LilyFirefly, line 525: I think we can do:

```suggestion
        let leading_whitespace = line.chars().take_while(char::is_whitespace).count();
```

src/render/filters.rs
```diff
@@ -502,6 +504,102 @@ impl ResolveFilter for WordcountFilter {
...
+
+            if current_len + word_len < width {
+                result.push(' ');
+                result.push_str(word);
+                current_len += 1 + word_len;
+            } else {
+                result.push('\n');
+                result.push_str(word);
+                current_len = word_len;
+            }
```
LilyFirefly, line 549: This saves a line, but is maybe not as clear. Up to you if you think it's better:

```suggestion
            if current_len + word_len < width {
                result.push(' ');
                current_len += 1 + word_len;
            } else {
                result.push('\n');
                current_len = word_len;
            }
            result.push_str(word);
```

src/render/filters.rs
```diff
@@ -502,6 +504,102 @@ impl ResolveFilter for WordcountFilter {
...
+        if let Some(bigint) = arg.to_bigint()
+            && let Some(n) = bigint.to_isize()
+            && n < 0
+        {
+            return Err(pyo3::exceptions::PyValueError::new_err(format!(
+                "invalid width {} (must be > 0)",
+                n
+            ))
+            .into());
+        }
```
LilyFirefly, line 586: I think we can merge the zero `width` check and error here:

```suggestion
        // Check for negative values before converting to usize
        if let Some(bigint) = arg.to_bigint()
            && let Some(n) = bigint.to_isize()
            && n <= 0
        {
            return Err(pyo3::exceptions::PyValueError::new_err(format!(
                "invalid width {} (must be > 0)",
                n
            ))
            .into());
        }
```

Then we don't need to check `width` in `wordwrap` and it doesn't need to return a `Result`.

tests/filters/test_wordwrap.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+def test_wordwrap_zero_width(assert_render_error):
+    assert_render_error(
+        template="{{ text|wordwrap:0 }}",
+        context={"text": "hello world"},
+        exception=ValueError,
+        django_message="invalid width 0 (must be > 0)",
+        rusty_message="invalid width 0 (must be > 0)",
+    )
```
LilyFirefly, line 266: We should use the `AnnotatePyErr` trait to improve the `rusty_message` in these tests.

tests/filters/test_wordwrap.py
```diff
@@ -0,0 +1,283 @@
...
+            "line one\nline two is longer",
+            "line one\nline two\nis longer",
+            id="preserves_line_breaks",
+        ),
+        pytest.param(10, "hello    world", "hello\nworld", id="multiple_spaces"),
+        pytest.param(10, "", "", id="empty_string"),
+        pytest.param(1, "a b c", "a\nb\nc", id="width_one"),
+        pytest.param(10, "hello world\n", "hello\nworld\n", id="trailing_newline"),
+        pytest.param(10, "   ", "   ", id="only_whitespace"),
+        pytest.param(10, "hello\tworld", "hello\nworld", id="tabs_and_spaces"),
```
LilyFirefly: This is just tabs, not tabs and spaces.

tests/filters/test_wordwrap.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+def test_wordwrap_with_autoescape_off(assert_render):
+    template = (
+        "{% autoescape off %}{{ a|wordwrap:3 }} {{ b|wordwrap:3 }}{% endautoescape %}"
+    )
+    assert_render(
+        template,
+        {"a": "a & b", "b": "a & b"},
+        "a &\nb a &\nb",
```
LilyFirefly: I don't think we need both `a` and `b` in the template/context here.

LilyFirefly (commented):
Amazing!

tests/filters/test_wordwrap.py
```diff
@@ -0,0 +1,283 @@
...
+    assert_render_error(
+        template="{{ text|wordwrap:0 }}",
+        context={"text": "hello world"},
+        exception=ValueError,
+        django_message="invalid width 0 (must be > 0)",
+        rusty_message="invalid width 0 (must be > 0)",
+    )
+
+
+@pytest.mark.skip("Should overflow in django")
```
LilyFirefly: Can you elaborate on why this needs to be skipped?

src/render/filters.rs
UnknownPlatypus: ye, much better. I was not really happy with the double nearly identical error message

src/render/filters.rs
UnknownPlatypus: Seems better (and will be [a clippy error](https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#branches_sharing_code) [soon](https://github.com/LilyFirefly/django-rusty-templates/pull/222))

src/render/filters.rs
UnknownPlatypus: It does not seem to be possible due to constraints on `take_while`

```
error[E0631]: type mismatch in function arguments
    --> src/render/filters.rs:532:58
     |
 532 |         let leading_whitespace = line.chars().take_while(char::is_whitespace).count();
     |                                               ---------- ^^^^^^^^^^^^^^^^^^^
     |                                               |          |
     |                                               |          expected due to this
     |                                               |          found signature defined here
     |                                               required by a bound introduced by this call
     |
     = note: expected function signature `for<'a> fn(&'a _) -> _`
                found function signature `fn(_) -> _`
note: required by a bound in `std::iter::Iterator::take_while`
    --> /home/thibaut/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/traits/iterator.rs:1201:12
     |
1198 |     fn take_while<P>(self, predicate: P) -> TakeWhile<Self, P>
     |        ---------- required by a bound in this associated function
...
1201 |         P: FnMut(&Self::Item) -> bool,
     |            ^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Iterator::take_while`
help: consider wrapping the function in a closure
     |
 532 |         let leading_whitespace = line.chars().take_while(|arg0: &char| char::is_whitespace(*arg0)).count();
     |                                                          +++++++++++++                    +++++++

```

tests/filters/test_wordwrap.py
UnknownPlatypus: my bad

tests/filters/test_wordwrap.py
UnknownPlatypus: Will add a better message 

I thought we would face similar OverFlow error and took tests from `test_center.py` but it appears only `str.center()` is failing with larger numbers. This issue is not with the `int(arg)` cast

wordwrap does not fail and only do comparison with the large integer

LilyFirefly: Do we run into any problems if we work with bigint? In the `center` case, the output string would become huge, but I don't think that applies here.

UnknownPlatypus: Not really, in this case, a very big arg value just mean the input string will be kept as-is.

But it also make no sense at all to use wordwrap with such value, just remove `|wordwrap` if you want a noop

LilyFirefly: I think this is another of those situations where it makes sense to seek a fix upstream in Django rather than diverge in API here.

I can also imagine a case where the `width` value is dynamic and set to something very high in some cases to set no wrapping. Of course it doesn't need to be as high as we're considering in this PR, but see my first point.

tests/filters/test_wordwrap.py
```diff
@@ -202,37 +204,63 @@ def test_wordwrap_invalid_width_string(assert_render_error):
 
 
 def test_wordwrap_negative_width(assert_render_error):
+    rusty_message = """\
+  × invalid width -5 (must be > 0)
+   ╭────
+ 1 │ {{ text|wordwrap:-5 }}
+   ·                  ─┬
+   ·                   ╰── argument
```
LilyFirefly: I'd call this `width` instead of `argument`.

UnknownPlatypus: I've changed that, however errors coming from other validations (like in `test_wordwrap_width_from_python_variable`) still have the other variant.

We might want to pass `argument_name` to some `RenderError` variant to improve this ?

tests/filters/test_wordwrap.py
UnknownPlatypus: > I think this is another of those situations where it makes sense to seek a fix upstream in Django rather than diverge in API here.

Wouldn't it better to fix the behavior here to align with upstream ? ie ensure we don't Overflow and just fallback to +inf if we fail to convert here ?

tests/filters/test_wordwrap.py
LilyFirefly: We could do, but I think the main thing I want is consistency between the first line of the error message and the tag pointing into the template source code. I think we now have that. I'm open to a PR though.

tests/filters/test_wordwrap.py
LilyFirefly: Sure, whatever makes us match Django's current behaviour.

LilyFirefly (approved):
Wonderful!

tests/filters/test_wordwrap.py
```diff
@@ -0,0 +1,362 @@
...
+        django_message="Python int too large to convert to C ssize_t",
+        rusty_message="""\
+  × Integer 9223372036854775808 is too large
+   ╭────
+ 1 │ {{ text|wordwrap:9223372036854775808 }}
+   ·                  ─────────┬─────────
+   ·                           ╰── here
+   ╰────
+""",
+    )
```
LilyFirefly: Ah, as we were discussing, I'd rather we match Django's behaviour here.

src/render/filters.rs
```diff
@@ -501,6 +503,99 @@ impl ResolveFilter for WordcountFilter {
...
+            ))
+            .annotate(py, self.argument.at, "width", template)
+            .into());
+        }
+
+        let width = match arg.resolve_usize(self.argument.at) {
+            Ok(w) => w,
+            Err(PyRenderError::RenderError(RenderError::OverflowError { .. })) => usize::MAX,
+            Err(e) => return Err(e),
+        };
```
LilyFirefly, line 592: I suspect we can optimise this logic a bit more, but it's probably not worth it.

LilyFirefly (approved):

src/render/filters.rs
UnknownPlatypus: I have a feeling `resolve_usize`  / `resolve_bigint` / `to_bigint` probably need consolidation.
It could have additional arguments controlling if negative are allowed, zero is allowed, overflow should be capped.

Then wrapper would create appropriate errors based on the context (are we converting a value, a filter argument, something else ...)

PR #227, opened 2025-11-28
feat: Implement TemplateContent for TagToken and TagParts

christophehenry:
There's a lot of `pub struct <SomeStruct> { pub at: (usize, usize) }` and they should all be implmenting `TemplateContent` by default. Is there a way to express that in Rust?


LilyFirefly:
Is there anywhere in Rusty Templates that can use these? Or do you just need it for django-template-transpiler? Either way, I'm happy to merge this once we have test coverage.

LilyFirefly (approved):
Great, thanks!

PR #229, opened 2025-11-29
chore: Move crate::types::Variable to dtl_lexer::types::Variable

christophehenry:
This is part of an effort to move more parts of tag parsing into the `dtls-lexer` crate as I'm beginning to implement tag rendering in django-template-transpiler. Starting with `Variable` and integrating it with `VariableTokenType` will allow me to support `{{ var.subvar.subsubvar }}` which DTT doesn't support yet.

### DIFF
--- a/dtl-lexer/src/variable.rs
+++ b/dtl-lexer/src/variable.rs
@@ -7,7 +7,7 @@ use crate::common::{
     LexerError, NextChar, check_variable_attrs, lex_numeric, lex_text, lex_translated,
     lex_variable_argument, trim_variable,
 };
-use crate::types::TemplateString;
+use crate::types::{TemplateString, Variable};
 use crate::{END_TRANSLATE_LEN, QUOTE_LEN, START_TRANSLATE_LEN, TemplateContent};
 
 #[derive(Debug, PartialEq, Eq)]
@@ -66,23 +66,10 @@ impl<'t> TemplateContent<'t> for FilterToken {
 }
 
 #[derive(Debug, PartialEq)]
-pub enum VariableTokenType {
-    Variable,
+pub enum VariableToken {
+    Variable(Variable),
     Int(BigInt),
     Float(f64),
-}
-
-#[derive(Debug, PartialEq)]
-pub struct VariableToken {
-    pub at: (usize, usize),
-    pub token_type: VariableTokenType,
-}
-
-impl<'t> TemplateContent<'t> for VariableToken {
-    fn content(&self, template: TemplateString<'t>) -> &'t str {
-        let (start, len) = self.at;
-        &template.0[start..start + len]
-    }
 }
 
 #[derive(Error, Debug, Diagnostic, PartialEq, Eq)]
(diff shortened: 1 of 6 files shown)


dtl-lexer/src/variable.rs
```diff
@@ -118,6 +118,9 @@ pub fn lex_variable(
 
     let start = start + variable.len() - rest.len();
     let content = trim_variable(rest);
+    let end = content.len();
+    let at = (start, end);
+
```
LilyFirefly: These can go below the `content.is_empty()` check (currently just below this comment).

src/parse.rs
```diff
@@ -1096,7 +1096,7 @@ impl<'t, 'l, 'py> Parser<'t, 'l, 'py> {
             return Err(ParseError::EmptyVariable { at: at.into() });
         };
         let mut var = match variable_token.token_type {
-            VariableTokenType::Variable => self.parse_for_variable(variable_token.at).into(),
+            VariableTokenType::Variable(_) => self.parse_for_variable(variable_token.at).into(),
```
LilyFirefly: We can get `at` from the `Variable` now:

```suggestion
            VariableTokenType::Variable(variable) => self.parse_for_variable(variable.at).into(),
```

dtl-lexer/src/variable.rs
```diff
@@ -67,7 +67,7 @@ impl<'t> TemplateContent<'t> for FilterToken {
 
 #[derive(Debug, PartialEq)]
 pub enum VariableTokenType {
```
LilyFirefly: I think we should be able to merge `VariableTokenType` and `VariableToken` now, since `at` is actually only used for the `VariableTokenType::Variable` variant.

LilyFirefly (commented):
This looks good to me, but we can simplify further I think.

That said, I haven't checked usage in django-template-transpiler, so maybe you still need things we don't here.

dtl-lexer/src/variable.rs
```diff
@@ -66,22 +67,19 @@ impl<'t> TemplateContent<'t> for FilterToken {
...
-
 impl<'t> TemplateContent<'t> for VariableToken {
     fn content(&self, template: TemplateString<'t>) -> &'t str {
-        let (start, len) = self.at;
-        &template.0[start..start + len]
+        match self {
+            VariableToken::Variable(var) => template.content(var.at),
+            VariableToken::Int(num) => num.to_string().as_str(),
+            VariableToken::Float(num) => num.to_string().as_str(),
+        }
```
christophehenry: @LilyFirefly So I integrated your review and merged `VariableTokenType` and `VariableToken` but I can't solve this one. I can't find how to produce a string with lifetime in Rust. On the other hand, I think this is used only in tests so I should be able to drop it.

LilyFirefly: Yeah, I think this can be dropped.

dtl-lexer/src/variable.rs
christophehenry: I was also thinking about doing the same with `Argument` and `ArgumentType` but from what I recall, I would have to modify [`lex_text`](https://github.com/LilyFirefly/django-rusty-templates/blob/main/dtl-lexer/src/common.rs#L79) and [`lex_translated`](https://github.com/LilyFirefly/django-rusty-templates/blob/main/dtl-lexer/src/common.rs#L107) that only return text bounds and not text content.

LilyFirefly: I'm happy for us to look into that, but in a separate PR.

dtl-lexer/src/variable.rs
```diff
@@ -388,13 +377,10 @@ mod tests {
...
-            token,
-            VariableToken {
-                at: (3, 7),
-                token_type: VariableTokenType::Variable
-            }
-        );
+        assert_eq!(token, VariableToken::Variable(Variable::new((3, 7))));
+        let VariableToken::Variable(token) = token else {
+            unreachable!()
+        };
```
christophehenry, line 383: IDK how to better write that.

PR #235, opened 2025-12-03
Add engine attribute to Template and Parser

LilyFirefly:
This uses an `Arc` to support cheaply sharing the `Engine` between multiple templates. The `Engine` also now uses a `Mutex` to provide interior mutability for the template loaders, which are allowed to store state for caching purposes.

For context, I think I will need this for implementing the `{% include %}` tag (at least). This also matches Django's own internal API.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -47,6 +47,7 @@ use dtl_lexer::variable::{
     VariableToken, lex_variable,
 };
 
+use crate::template::django_rusty_templates::Engine;
 use crate::types::Argument;
 use crate::types::ArgumentType;
 use crate::types::ForVariable;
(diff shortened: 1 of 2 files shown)


src/loaders.rs
```diff
@@ -82,7 +83,7 @@ impl FileSystemLoader {
         &self,
         py: Python<'_>,
         template_name: &str,
-        engine: &EngineData,
+        engine: Arc<Engine>,
```
UnknownPlatypus, line 86: I suppose the thread safety of `Arc` is required and we cannot use `Rc` that is a bit cheaper ?

LilyFirefly: Yeah, we need thread safety because Python can send things between threads behind our back.

PR #236, opened 2025-12-03
feat: merge ArgumentType into Argument

christophehenry:
As proposed in #229, this PR merges ArgumentType into Argument to simplify the lexer API.

### DIFF
--- a/dtl-lexer/src/common.rs
+++ b/dtl-lexer/src/common.rs
@@ -80,18 +80,19 @@ pub fn lex_variable(byte: usize, rest: &str) -> (At, usize, &str) {
 pub fn lex_text<'t>(
     byte: usize,
     rest: &'t str,
-    chars: &mut std::str::Chars,
+    token: &'t str,
     end: char,
-) -> Result<(At, usize, &'t str), LexerError> {
+) -> Result<(At, &'t str, usize, &'t str), LexerError> {
     let mut count = 1;
+    let mut chars = token.chars().enumerate();
     loop {
-        let Some(next) = chars.next() else {
+        let Some((_, next)) = chars.next() else {
             let at = (byte, count);
             return Err(LexerError::IncompleteString { at: at.into() });
         };
         count += next.len_utf8();
         if next == '\\' {
-            let Some(next) = chars.next() else {
+            let Some((_, next)) = chars.next() else {
                 let at = (byte, count);
                 return Err(LexerError::IncompleteString { at: at.into() });
             };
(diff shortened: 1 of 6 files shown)


dtl-lexer/src/variable.rs
```diff
@@ -354,7 +308,18 @@ mod tests {
...
+                            (t.content(template), Some(num.to_string().as_str()))
+                        }
+                        Argument::Text(text) => (t.content(template), Some(text.as_str())),
+                        Argument::TranslatedText(text) => {
+                            (t.content(template), Some(text.as_str()))
+                        }
+                        Argument::Variable(var) => {
+                            (t.content(template), Some(template.content(var.at)))
+                        }
+                    },
```
christophehenry, line 311: So this creates a compilation error. As I understand, I return a `&str` that is deallocated at the end of the scope but I don't know how to solve this. There seems to be no way to force a lifetime onto a pointer. I'm curious how to solve this.

dtl-lexer/src/common.rs
```diff
@@ -147,7 +149,12 @@ pub fn lex_numeric(byte: usize, rest: &str) -> ((usize, usize), usize, &str) {
     };
     // End match django bug
     let at = (byte, end);
-    (at, byte + end, &rest[end..])
+    (
+        at,
+        rest[byte..end].parse().unwrap(),
```
LilyFirefly: This is a non-starter for me for two reasons:

* This uses `unwrap`. This function would need to return a `Result` instead and we'd need to track that error through.
* This is doing parsing. That's deliberately a different step in the architecture here.

dtl-lexer/src/common.rs
```diff
@@ -81,14 +81,16 @@ pub fn lex_text<'t>(
     rest: &'t str,
     chars: &mut std::str::Chars,
     end: char,
-) -> Result<((usize, usize), usize, &'t str), LexerError> {
+) -> Result<((usize, usize), String, usize, &'t str), LexerError> {
     let mut count = 1;
+    let mut text = String::new();
```
LilyFirefly: This is doing a memory allocation we didn't need before. Previously we were able to obtain the same string by slicing into `template` by the `(usize, usize)` we're returning. Using an owned `String` is also causing your lifetime issues you called out. Previously the `template` slice automatically has the right lifetime - that of `template`.

dtl-lexer/src/variable.rs
LilyFirefly: You've rediscovered a reason why we use the current design of returning a `(usize, usize)` (`at`) in the lexer stage. This avoids needing to store references with lifetimes or needing extra allocations for owned values.

LilyFirefly (commented):
Thanks for looking into this!

However, I strongly suspect we can't refactor this for the reasons I mention below. Let me know if there's anything I haven't explained well.

dtl-lexer/src/common.rs
christophehenry: True but so does [`lex_variable`](https://github.com/LilyFirefly/django-rusty-templates/blob/main/dtl-lexer%2Fsrc%2Fvariable.rs#L118-L120), right? Anyway, I'm fine leaving this parsing on DTT part.

dtl-lexer/src/variable.rs
christophehenry: Ok then I think I can find a solution if, instead of passing `Chars` to `lex_*`, it requires an `&'t str` slice?

dtl-lexer/src/common.rs
LilyFirefly: Good point about `lex_variable`, though I'm not entirely sure I like that design.

christophehenry:
Ok, so this is the essence of my refactor. It still can be simplified further, though since [`lex_text` now takes 2 arguments that are mostly the same slice except for one item](https://github.com/LilyFirefly/django-rusty-templates/pull/236/files#diff-34a9e6066c748622fd98dac4deb3de6b40c605965f93b3d10476758cbd58eaa3R80-R83).

A few things seems to be broken on the parser side that seem mostly that I unintentionnally changed how quotes are handled. I still need to figure out that, which is difficult since PyO3 doesn't want to detect Django on my system event if it is installed globally.

All in all the spirit of this refactor is there and should be completable but since it's breaking most of the lexer API, I'd like an opinion before I spend a few more hours fixing the tests.

dtl-lexer/src/variable.rs
```diff
@@ -3,62 +3,28 @@ use num_bigint::BigInt;
...
-impl<'t> TemplateContent<'t> for Argument {
-    fn content(&self, template: TemplateString<'t>) -> &'t str {
-        let (start, len) = self.content_at();
-        &template.0[start..start + len]
-    }
+#[derive(Debug, PartialEq)]
+pub enum Argument<'t> {
+    Numeric(At, &'t str),
+    Text(At, &'t str),
+    TranslatedText(At, &'t str),
```
christophehenry, line 17: So I did remove the parsing here. Now the values are untyped and represented as `str`. I think `dlt_lexer::types::ArgumentType` should be changed to follow the same rule at some point.

PR #237, opened 2025-12-03
Implement the {% include %} template tag

LilyFirefly:
Part of #2.

### DIFF
--- a/dtl-lexer/src/tag/custom_tag.rs
+++ b/dtl-lexer/src/tag/custom_tag.rs
@@ -34,6 +34,14 @@ impl SimpleTagToken {
         }
     }
 
+    /// The location of the full keyword argument:
+    ///     keyword=variable
+    ///             ────┬───
+    ///                 ╰── self.at
+    ///     ───┬───
+    ///        ╰── self.kwarg
+    ///     ────────┬───────
+    ///             ╰── self.all_at()
     pub fn all_at(&self) -> At {
         match self.kwarg {
             None => self.at,


src/parse.rs
```diff
@@ -105,6 +108,7 @@ fn unexpected_argument(filter: &'static str, right: Argument) -> ParseError {
 #[derive(Clone, Debug, PartialEq)]
 pub struct Filter {
     pub at: (usize, usize),
+    pub all_at: (usize, usize),
```
LilyFirefly: I don't really like this name, but I haven't come up with a better one yet.

UnknownPlatypus:
Will review tomorrow when I get some time, but I just noticed that some include tests are still failing in the django check, dunno if it’s beeing flaky or actual issues but you might want to check 
https://github.com/LilyFirefly/django-rusty-templates/pull/237#issuecomment-3613577910

LilyFirefly:
> Will review tomorrow when I get some time, but I just noticed that some include tests are still failing in the django check, dunno if it’s beeing flaky or actual issues but you might want to check [#237 (comment)](https://github.com/LilyFirefly/django-rusty-templates/pull/237#issuecomment-3613577910)

`test_include10` through to `test_include_14` all fail because we don't support `Engine.string_if_invalid` yet.
`test_extends_include_missing_baseloader` and `test_extends_include_missing_cacheloader` both require the `{% extends %}` tag.

I haven't worked out why `test_include_cache` fails, but that looks more like a bug in custom template tags than in `include`.

`test_include04`, `test_include_template_iterable` and `IncludeRelativeBehaviorTests.test_include_error` each pass locally and I haven't worked out why they fail in CI.

dtl-lexer/src/tag/custom_tag.rs
```diff
@@ -33,6 +33,13 @@ impl SimpleTagToken {
             SimpleTagTokenType::TranslatedText => translated_text_content_at(self.at),
         }
     }
+
+    pub fn all_at(&self) -> At {
+        match self.kwarg {
+            None => self.at,
+            Some(kwarg_at) => (kwarg_at.0, self.at.0 - kwarg_at.0 + self.at.1),
+        }
+    }
```
UnknownPlatypus, line 50: Maybe just my grug brain but I'm strugly to understand what this represent. Could we add a doc-comment maybe with an exemple?

dtl-lexer/src/tag/include.rs
```diff
@@ -0,0 +1,193 @@
...
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum IncludeTemplateTokenType {
+    Text,
+    Variable,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct IncludeTemplateToken {
+    pub at: (usize, usize),
```
UnknownPlatypus:
```suggestion
    pub at: At,
```
I suppose it should be that ?

We should maybe add a simple regex pre-commit rule to catch these ?

UnknownPlatypus: Something like that would work:

```yaml
  - repo: local
    hooks:
      - id: lexer-use-at
        name: "Replace `(usize, usize)` with `crate::types::At`"
        language: pygrep
        entry: \(usize, usize\)
        types: [rust]
        exclude: "dtl-lexer/src/types.rs"
```

scripts/django_tests_use_rusty_templates.patch
```diff
@@ -27,3 +27,163 @@ index ccfc327..1ab9ace 100755
...
++            template.render({})
+ 
+     def test_extends_include_missing_cachedloader(self):
+         engine = Engine(
+@@ -266,23 +263,21 @@ class IncludeTests(SimpleTestCase):
+ 
+         template = engine.get_template("test_extends_error.html")
+         with self.assertRaisesMessage(TemplateDoesNotExist, "missing.html"):
+-            template.render(Context())
++            template.render({})
```
UnknownPlatypus, line 106: Should we gracefully handle if a Context object is passed instead of a dict ?

LilyFirefly: Maybe, I'm not sure.

dtl-lexer/src/tag/include.rs
LilyFirefly: We could, but I think this one slipped in only because I was working on this branch before we introduced the `At` alias.

dtl-lexer/src/tag/custom_tag.rs
LilyFirefly: I've added a docstring, which I hope helps!

scripts/django_tests_use_rusty_templates.patch
LilyFirefly: It depends whether we want to only support the ["high level, backend agnostic APIs"](https://docs.djangoproject.com/en/6.0/ref/templates/api/#overview) or if we want to support the low level DTL-specific API. I think `Context` is part of the DTL-specific API.

UnknownPlatypus: Either is fine I suppose, this just make the patch a bit bigger but maybe its fine to avoid implementing the internal dtl handling.
Up to you

dtl-lexer/src/tag/custom_tag.rs
UnknownPlatypus: Thanks! Makes sense

scripts/django_tests_use_rusty_templates.patch
LilyFirefly: I won't add it now, but we can always revisit later.

PR #242, opened 2025-12-05
Refactor lex_variable

LilyFirefly:
Rename to `lex_variable_or_filter` to better distinguish it from `lex_variable` which is used to lex the dotted lookup syntax for variables.

Return an `at` location from `lex_variable_or_filter` instead of only providing it for `VariableToken::Variable`. This is useful in #237 for error reporting in `{% include %}` tags.

### DIFF
--- a/dtl-lexer/src/variable.rs
+++ b/dtl-lexer/src/variable.rs
@@ -7,7 +7,7 @@ use crate::common::{
     LexerError, NextChar, check_variable_attrs, lex_numeric, lex_text, lex_translated,
     lex_variable_argument, trim_variable,
 };
-use crate::types::TemplateString;
+use crate::types::{At, TemplateString};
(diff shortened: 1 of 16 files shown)


dtl-lexer/src/variable.rs
```diff
@@ -94,10 +94,12 @@ pub enum VariableLexerError {
     },
 }
 
-pub fn lex_variable(
+type At = (usize, usize);
```
UnknownPlatypus: Should we plan to reuse this type alias in the rest of the crate ?
I'm also wondering if it would make sense to have an helper to convert from this to `miette::SourceSpan` for error reporting ?

LilyFirefly: Yeah, I had this idea in the back of my mind. For the conversion to `SourceSpan`, do you have anything in mind other than `.into()`, which already works?

UnknownPlatypus: Oh no, if into works thats great

christophehenry (approved):

PR #261, opened 2025-12-21
Implement `cut` builtin filter as part of #1

p-r-a-v-i-n:
part of #1

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -18,6 +18,7 @@ use crate::filters::AddFilter;
 use crate::filters::AddSlashesFilter;
 use crate::filters::CapfirstFilter;
 use crate::filters::CenterFilter;
+use crate::filters::CutFilter;
 use crate::filters::DefaultFilter;
 use crate::filters::DefaultIfNoneFilter;
 use crate::filters::EscapeFilter;
@@ -136,6 +137,10 @@ impl Filter {
             },
             "center" => match right {
                 Some(right) => FilterType::Center(CenterFilter::new(right)),
+                None => return Err(ParseError::MissingArgument { at: at.into() }),
+            },
+            "cut" => match right {
+                Some(right) => FilterType::Cut(CutFilter::new(right)),
                 None => return Err(ParseError::MissingArgument { at: at.into() }),
             },
             "default" => match right {
--- a/src/render/filters.rs
+++ b/src/render/filters.rs
@@ -9,10 +9,10 @@ use pyo3::types::PyType;
 
 use crate::error::{AnnotatePyErr, PyRenderError, RenderError};
 use crate::filters::{
-    AddFilter, AddSlashesFilter, CapfirstFilter, CenterFilter, DefaultFilter, DefaultIfNoneFilter,
-    EscapeFilter, EscapejsFilter, ExternalFilter, FilterType, LengthFilter, LowerFilter,
-    SafeFilter, SlugifyFilter, TitleFilter, UpperFilter, WordcountFilter, WordwrapFilter,
-    YesnoFilter,
+    AddFilter, AddSlashesFilter, CapfirstFilter, CenterFilter, CutFilter, DefaultFilter,
+    DefaultIfNoneFilter, EscapeFilter, EscapejsFilter, ExternalFilter, FilterType, LengthFilter,
+    LowerFilter, SafeFilter, SlugifyFilter, TitleFilter, UpperFilter, WordcountFilter,
+    WordwrapFilter, YesnoFilter,
 };
 use crate::parse::Filter;
 use crate::render::common::gettext;
@@ -46,6 +46,7 @@ impl Resolve for Filter {
             FilterType::AddSlashes(filter) => filter.resolve(left, py, template, context),
             FilterType::Capfirst(filter) => filter.resolve(left, py, template, context),
             FilterType::Center(filter) => filter.resolve(left, py, template, context),
+            FilterType::Cut(filter) => filter.resolve(left, py, template, context),
             FilterType::Default(filter) => filter.resolve(left, py, template, context),
             FilterType::DefaultIfNone(filter) => filter.resolve(left, py, template, context),
             FilterType::Escape(filter) => filter.resolve(left, py, template, context),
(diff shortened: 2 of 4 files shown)


src/parse.rs
```diff
@@ -137,6 +138,10 @@ impl Filter {
                 Some(right) => FilterType::Center(CenterFilter::new(right)),
                 None => return Err(ParseError::MissingArgument { at: at.into() }),
             },
+            "cut" => match right {
+                Some(right) => FilterType::Cut(CutFilter::new(right)),
+                None => return Err(ParseError::MissingArgument { at: at.into() }),
```
LilyFirefly, line 144: We'll need a test for this line. There should be good examples for other filters you can copy.

src/render/filters.rs
```diff
@@ -189,6 +191,46 @@ impl ResolveFilter for CenterFilter {
...
+impl ResolveFilter for CutFilter {
+    fn resolve<'t, 'py>(
+        &self,
+        variable: Option<Content<'t, 'py>>,
+        py: Python<'py>,
+        template: TemplateString<'t>,
+        context: &mut Context,
+    ) -> ResolveResult<'t, 'py> {
+        let Some(variable) = variable else {
+            return Ok(Some("".as_content()));
```
LilyFirefly, line 208: This will need a test too.

src/render/filters.rs
```diff
@@ -189,6 +191,46 @@ impl ResolveFilter for CenterFilter {
...
+
+        let arg = self
+            .argument
+            .resolve(py, template, context, ResolveFailures::Raise)?
+            .expect("cut filter requires an argument")
+            .resolve_string_strict(context, self.argument.at.into())?
+            .as_raw()
+            .clone();
+
+        let content_string = variable.resolve_string(context)?;
```
LilyFirefly, line 218: If you can add tests for the error paths here (the `?`), that'd be great too.

src/render/filters.rs
```diff
@@ -189,6 +191,46 @@ impl ResolveFilter for CenterFilter {
...
+        context: &mut Context,
+    ) -> ResolveResult<'t, 'py> {
+        let Some(variable) = variable else {
+            return Ok(Some("".as_content()));
+        };
+
+        let arg = self
+            .argument
+            .resolve(py, template, context, ResolveFailures::Raise)?
+            .expect("cut filter requires an argument")
```
LilyFirefly: I think this message (that we also use for `CenterFilter`) is more accurate as to why it's safe to unwrap the `Option` here:

```suggestion
            .expect("missing argument in context should already have raised")
```

src/render/filters.rs
```diff
@@ -189,6 +191,46 @@ impl ResolveFilter for CenterFilter {
...
+            .resolve_string_strict(context, self.argument.at.into())?
+            .as_raw()
+            .clone();
```
LilyFirefly: I think we can avoid the `clone` with `into_raw`:

```suggestion
            .into_raw();
```

src/render/filters.rs
```diff
@@ -189,6 +191,46 @@ impl ResolveFilter for CenterFilter {
...
+                if arg.as_ref() == ";" {
+                    ContentString::HtmlUnsafe(s.replace(arg.as_ref(), "").into())
+                } else {
+                    ContentString::HtmlSafe(s.replace(arg.as_ref(), "").into())
+                }
+            }
+            ContentString::HtmlUnsafe(s) => {
+                ContentString::HtmlUnsafe(s.replace(arg.as_ref(), "").into())
+            }
+        };
```
LilyFirefly, line 229: It might be nice to introduce a `cut` helper function:

```rust
fn cut(source: Cow<str>, value: &str) -> Cow<str> {
    source.replace(value, "").into()
}
```

Then this can become:

```suggestion
        let arg = arg.as_ref();
        let result = match content_string {
            ContentString::String(s) => ContentString::String(cut(s, arg)),
            ContentString::HtmlSafe(s) => {
                match arg {
                    ";" => ContentString::HtmlUnsafe(cut(s, arg)),
                    _ => ContentString::HtmlSafe(cut(s, arg)),
                }
            }
            ContentString::HtmlUnsafe(s) => ContentString::HtmlUnsafe(cut(s, arg)),
        };
```

p-r-a-v-i-n: I really like this suggestions

src/render/filters.rs
```diff
@@ -191,6 +192,48 @@ impl ResolveFilter for CenterFilter {
...
+        let content_string = variable.resolve_string(context)?;
+
+        let result = match content_string {
+            ContentString::String(s) => ContentString::String(cut(s, &arg_string).into()),
+            ContentString::HtmlSafe(s) => {
+                let cut = cut(s, &arg_string);
+                match arg {
+                    Cow::Borrowed(";") => ContentString::HtmlUnsafe(cut.into()),
+                    _ => ContentString::HtmlSafe(cut.into()),
+                }
```
LilyFirefly: I think we want this, so we also handle `Cow::Owned(";".to_string())`:

```suggestion
                match arg.as_ref() {
                    ";" => ContentString::HtmlUnsafe(cut.into()),
                    _ => ContentString::HtmlSafe(cut.into()),
                }
```

(https://doc.rust-lang.org/std/borrow/enum.Cow.html#method.as_ref)

src/render/filters.rs
```diff
@@ -191,6 +192,48 @@ impl ResolveFilter for CenterFilter {
...
+            .resolve_string_strict(context, self.argument.at.into())?
+            .into_raw();
+
+        let arg_string = arg.to_string();
+        let content_string = variable.resolve_string(context)?;
+
+        let result = match content_string {
+            ContentString::String(s) => ContentString::String(cut(s, &arg_string).into()),
+            ContentString::HtmlSafe(s) => {
+                let cut = cut(s, &arg_string);
```
LilyFirefly: We don't need the `arg_string` intermediate value:

```suggestion
        let content_string = variable.resolve_string(context)?;

        let result = match content_string {
            ContentString::String(s) => ContentString::String(cut(s, &arg).into()),
            ContentString::HtmlSafe(s) => {
                let cut = cut(s, &arg);
```

This way is also a bit more efficient because it doesn't need to allocate the `arg_string`.

src/render/filters.rs
```diff
@@ -191,6 +192,48 @@ impl ResolveFilter for CenterFilter {
...
+                match arg {
+                    Cow::Borrowed(";") => ContentString::HtmlUnsafe(cut.into()),
+                    _ => ContentString::HtmlSafe(cut.into()),
+                }
+            }
+            ContentString::HtmlUnsafe(s) => ContentString::HtmlUnsafe(cut(s, &arg_string).into()),
```
```suggestion
            ContentString::HtmlUnsafe(s) => ContentString::HtmlUnsafe(cut(s, &arg).into()),
```

LilyFirefly (approved):
Great! Thanks for applying those suggestions!

PR #263, opened 2025-12-22
Add length filter

p-r-a-v-i-n:
part #1

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -24,6 +24,7 @@ use crate::filters::EscapeFilter;
 use crate::filters::EscapejsFilter;
 use crate::filters::ExternalFilter;
 use crate::filters::FilterType;
+use crate::filters::LengthFilter;
 use crate::filters::LowerFilter;
 use crate::filters::SafeFilter;
 use crate::filters::SlugifyFilter;
@@ -156,6 +157,10 @@ impl Filter {
             "lower" => match right {
                 Some(right) => return Err(unexpected_argument("lower", right)),
                 None => FilterType::Lower(LowerFilter),
+            },
+            "length" => match right {
+                Some(right) => return Err(unexpected_argument("length", right)),
+                None => FilterType::Length(LengthFilter),
             },
             "safe" => match right {
                 Some(right) => return Err(unexpected_argument("safe", right)),
--- a/src/render/filters.rs
+++ b/src/render/filters.rs
@@ -10,8 +10,9 @@ use pyo3::types::PyType;
 use crate::error::{AnnotatePyErr, PyRenderError, RenderError};
 use crate::filters::{
     AddFilter, AddSlashesFilter, CapfirstFilter, CenterFilter, DefaultFilter, DefaultIfNoneFilter,
-    EscapeFilter, EscapejsFilter, ExternalFilter, FilterType, LowerFilter, SafeFilter,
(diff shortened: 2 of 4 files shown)


src/render/filters.rs
```diff
@@ -346,6 +348,25 @@ impl ResolveFilter for LowerFilter {
...
+    fn resolve<'t, 'py>(
+        &self,
+        variable: Option<Content<'t, 'py>>,
+        _py: Python<'py>,
+        _template: TemplateString<'t>,
+        _context: &mut Context,
+    ) -> ResolveResult<'t, 'py> {
+        let len = match variable {
+            None => 0,
+            Some(Content::String(s)) => s.as_raw().chars().count(),
```
LilyFirefly, line 367: It would be good to have a test with a non-ascii string. Emoji would be a good choice.

```python
def test_length_unicode(assert_render):
    template = "{{ var|lower|length }}"
    assert_render(template=template, context={"var": "🐍"}, expected="1")
```

I've included the `lower` filter to make sure we're working with `Content::String`, not `Content::Py`. We should also have a version of the test without the `lower` filter.

src/parse.rs
```diff
@@ -157,6 +158,10 @@ impl Filter {
                 Some(right) => return Err(unexpected_argument("lower", right)),
                 None => FilterType::Lower(LowerFilter),
             },
+            "length" => match right {
+                Some(right) => return Err(unexpected_argument("length", right)),
```
LilyFirefly, line 162: We'll need a test for this error.

src/render/filters.rs
```diff
@@ -346,6 +348,25 @@ impl ResolveFilter for LowerFilter {
...
+impl ResolveFilter for LengthFilter {
+    fn resolve<'t, 'py>(
+        &self,
+        variable: Option<Content<'t, 'py>>,
+        _py: Python<'py>,
+        _template: TemplateString<'t>,
+        _context: &mut Context,
+    ) -> ResolveResult<'t, 'py> {
+        let len = match variable {
+            None => 0,
```
LilyFirefly, line 366: The `None` variant of `variable` happens when the variable isn't in the context, so to trigger this path, we'll need a test like:

```python
def test_length_undefined(assert_render):
    template = "{{ var|length }}"
    assert_render(template=template, context={}, expected="0")
```

src/render/filters.rs
```diff
@@ -346,6 +348,25 @@ impl ResolveFilter for LowerFilter {
...
+            Some(Content::String(s)) => s.as_raw().chars().count(),
+            Some(Content::Py(obj)) => obj.len().unwrap_or(0),
+            Some(Content::Int(_) | Content::Float(_) | Content::Bool(_)) => 0,
```
LilyFirefly, line 369: We'll also need tests for these. Int and float can just use literals in the template, but for bool we'll need something like:

```python
def test_length_chained_with_bool(assert_render):
    template = "{% for x in 'ab' %}{{ forloop.first|length }}{% endfor %}"
    assert_render(template=template, context={}, expected="00")
```

src/render/filters.rs
p-r-a-v-i-n: Resolved.

LilyFirefly (approved):
Amazing, thanks!

PR #267, opened 2025-12-23
Add missing tests for `capfirst` builtin filter

p-r-a-v-i-n:
I think there test for https://github.com/LilyFirefly/django-rusty-templates/blob/cc63cb68dd84cbd16ed0128df385199019c6df85/src/parse.rs#L133 was missing , i have added test for it.


LilyFirefly (approved):
Thanks!

PR #268, opened 2025-12-23
Add helper function for generating rusty error messages

p-r-a-v-i-n:
For me Currently, it's PITA to write Rust error messages. To do so, I first have to run pytest, check the format of the error, and then copy-paste the message. Even though I’ve fixed the indentation multiple times, it still doesn’t work as expected.

So, I’ve added a helper function to fix this.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -19,3 +19,30 @@ class BrokenDunderHtml(str):
 class BrokenDunderBool:
     def __bool__(self):
         1 / 0
+
+
+def get_rusty_error(template, header, target, sub_label="here"):
+    """
+    Automated Miette-style error generator.
+    """
+    offset = template.find(target)
+    if offset == -1:
+        offset = 0
+
+    width = len(target)
+
+    side_dashes_count = (width - 1) // 2
+    side_dashes = "─" * side_dashes_count
+    pointer = f"{side_dashes}┬{side_dashes}"
+
+    label_padding = " " * (offset + side_dashes_count)
+    padding = " " * offset
+
+    return f"""\
+  × {header}
+   ╭────
+ 1 │ {template}
+   · {padding}{pointer}
+   · {label_padding}╰── {sub_label}
+   ╰────
+"""


p-r-a-v-i-n:
instead of this. 
  
>
```
    rusty_message = """\
  × wordcount filter does not take an argument
   ╭────
 1 │ {{ var|wordcount:arg }}
   ·                  ─┬─
   ·                   ╰── unexpected argument
   ╰────
"""
"""
```

i can use it like this

>  template = "{{ var|wordcount:arg }}"
`rusty_message = get_rusty_error(template, "wordcount filter does not take an argument", "unexpected argument")`

LilyFirefly:
> i'm also doing this but I struggle.

I'm sorry it's difficult.

UnknownPlatypus:
I think I found a neat solution with https://github.com/15r10nk/inline-snapshot.

For ex with this test:

```python
from inline_snapshot import snapshot
def test_default_broken_bool(assert_render_error):
    django_message = snapshot()
    rusty_message = snapshot()
    assert_render_error(
        template="{{ value|default:'default' }}",
        context={"value": Broken()},
        exception=ZeroDivisionError,
        django_message=django_message,
        rusty_message=rusty_message,
    )
```
You can do `pytest -k test_default_broken_bool --inline-snapshot=create` and you get this:

```python
def test_default_broken_bool(assert_render_error):
    django_message = snapshot("division by zero")
    rusty_message = snapshot("""\
  × division by zero
   ╭────
 1 │ {{ value|default:'default' }}
   ·          ───┬───
   ·             ╰── here
   ╰────
""")
    assert_render_error(
        template="{{ value|default:'default' }}",
        context={"value": Broken()},
        exception=ZeroDivisionError,
        django_message=django_message,
        rusty_message=rusty_message,
    )
```

Doing updates seem to be as simple as doing `--inline-snapshot=fix` or `--inline-snapshot=review` (if you want a prompt with the diff first.)

Should I send a pr to switch to this ?

LilyFirefly:
Ooh!! Great idea! I've used inline-snapshot before and it works well. I'm a bit annoyed I didn't think of it myself!

PR #269, opened 2025-12-23
Add initial codspeed setup

UnknownPlatypus:
You'll need to register on codespeed for the workflow to run.

I'm not super fan of having to make some stuff pub for the bench to catch them.
Do you know of another way ?


UnknownPlatypus:
> Do you think the cargo-shear lint failures need fixing? I'm happy to merge otherwise.

I think it's correct, I've fixed the issue. 
Seems to be [a day in y](https://downdetector.com/status/github/) at github however, getting random 504 in ci

PR #270, opened 2025-12-24
Add date filter

p-r-a-v-i-n:
part of #1

### DIFF
--- a/src/render/filters.rs
+++ b/src/render/filters.rs
@@ -301,11 +301,11 @@ impl ResolveFilter for DateFilter {
             }
         };
 
-        let value_py = value.to_py(py);
+        let value = value.to_py(py);
 
-        let is_valid = value_py.is_instance_of::<PyDate>()
-            || value_py.is_instance_of::<PyTime>()
-            || value_py.is_instance_of::<PyDateTime>();
+        let is_valid = value.is_instance_of::<PyDate>()
+            || value.is_instance_of::<PyTime>()
+            || value.is_instance_of::<PyDateTime>();
 
         if !is_valid {
             return Ok(Some("".as_content()));
@@ -313,13 +313,15 @@ impl ResolveFilter for DateFilter {
 
         let date_format_fn = DATE_FORMAT.import(py, "django.utils.dateformat", "format")?;
 
-        let formatted = match date_format_fn.call1((value_py, fmt)) {
+        let formatted = match date_format_fn.call1((value, fmt)) {
             Ok(res) => res,
             Err(e) => {
                 if e.is_instance_of::<pyo3::exceptions::PyAttributeError>(py) {
                     return Ok(Some("".as_content()));
                 }
-                return Err(PyRenderError::PyErr(e));
+                return Err(PyRenderError::PyErr(
+                    e.annotate(py, self.at, "here", template),
+                ));
             }
         };
 
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -94,11 +94,12 @@ impl DefaultIfNoneFilter {
 #[derive(Clone, Debug, PartialEq)]
 pub struct DateFilter {
     pub argument: Option<Argument>,
+    pub at: At,
 }
 
 impl DateFilter {
-    pub fn new(argument: Option<Argument>) -> Self {
-        Self { argument }
+    pub fn new(argument: Option<Argument>, at: At) -> Self {
+        Self { argument, at }
     }
 }
 
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -158,7 +158,7 @@ impl Filter {
                 Some(right) => FilterType::DefaultIfNone(DefaultIfNoneFilter::new(right)),
                 None => return Err(ParseError::MissingArgument { at: at.into() }),
             },
-            "date" => FilterType::Date(DateFilter::new(right)),
+            "date" => FilterType::Date(DateFilter::new(right, at)),
             "escape" => match right {
                 Some(right) => return Err(unexpected_argument("escape", right)),
                 None => FilterType::Escape(EscapeFilter),


src/render/filters.rs
```diff
@@ -274,6 +275,51 @@ impl ResolveFilter for DefaultIfNoneFilter {
...
+        };
+
+        let fmt = match &self.argument {
+            Some(arg) => arg
+                .resolve(py, template, context, ResolveFailures::Raise)?
+                .expect("missing date format argument")
+                .to_py(py),
+            None => {
+                let formats_mod = py.import("django.utils.formats")?;
+                formats_mod.call_method1("get_format", ("DATE_FORMAT",))?
```
LilyFirefly: Instead of using `py.import` we should use the [more efficient approach with `PyOnceLock`](https://pyo3.rs/main/doc/pyo3/sync/struct.pyoncelock#method.import).

See also #244.

src/render/filters.rs
```diff
@@ -274,6 +275,51 @@ impl ResolveFilter for DefaultIfNoneFilter {
...
+        context: &mut Context,
+    ) -> ResolveResult<'t, 'py> {
+        let Some(value) = variable else {
+            return Ok(Some("".as_content()));
+        };
+
+        let fmt = match &self.argument {
+            Some(arg) => arg
+                .resolve(py, template, context, ResolveFailures::Raise)?
+                .expect("missing date format argument")
```
LilyFirefly: We should explain why it's OK to use `expect`:

```suggestion
                .expect("missing argument in context should already have raised")
```

src/render/filters.rs
```diff
@@ -274,6 +275,51 @@ impl ResolveFilter for DefaultIfNoneFilter {
...
+            None => {
+                let formats_mod = py.import("django.utils.formats")?;
+                formats_mod.call_method1("get_format", ("DATE_FORMAT",))?
+            }
+        };
+
+        let value_py = value.to_py(py);
+        let is_date = py.import("datetime")?.getattr("date")?;
+        let is_time = py.import("datetime")?.getattr("time")?;
+        if !value_py.is_instance(&is_date)? && !value_py.is_instance(&is_time)? {
```
LilyFirefly: We might be able to use [`PyDate`](https://pyo3.rs/main/doc/pyo3/types/struct.pydatehttps://pyo3.rs/main/doc/pyo3/types/struct.pydate) and [`PyTime`](https://pyo3.rs/main/doc/pyo3/types/struct.pytime) here, which could avoid needing `import` and `getattr`.

Also, should we handle `datetime` too, in addition to `date` and `time`?

tests/filters/test_date.py
```diff
@@ -0,0 +1,203 @@
...
+    )
+
+
+def test_date_function():
+    from django.template.defaultfilters import date
+
+    assert date(datetime(2005, 12, 29), "d F Y") == "29 December 2005"
+    assert date("") == ""
+    assert date(None) == ""
+    assert date(datetime(2005, 12, 29), r"jS \o\f F") == "29th of December"
```
LilyFirefly: We don't need to test Django's code directly unless we're comparing it to how our Rust code works.

p-r-a-v-i-n: I'm sorry about this part.

LilyFirefly: No need to apologise! :smile:

src/render/filters.rs
```diff
@@ -274,6 +278,55 @@ impl ResolveFilter for DefaultIfNoneFilter {
...
+                let get_format = GET_FORMAT.import(py, "django.utils.formats", "get_format")?;
+                get_format.call1(("DATE_FORMAT",))?
+            }
+        };
+
+        let value_py = value.to_py(py);
+
+        let is_valid = value_py.is_instance_of::<PyDate>()
+            || value_py.is_instance_of::<PyTime>()
+            || value_py.is_instance_of::<PyDateTime>();
```
LilyFirefly: Much nicer!

src/render/filters.rs
```diff
@@ -274,6 +278,55 @@ impl ResolveFilter for DefaultIfNoneFilter {
...
+
+        if !is_valid {
+            return Ok(Some("".as_content()));
+        }
+
+        let date_format_fn = DATE_FORMAT.import(py, "django.utils.dateformat", "format")?;
+
+        let formatted = match date_format_fn.call1((value_py, fmt)) {
```
LilyFirefly: We can reuse names in rust:

```suggestion
        let valuu = value.to_py(py);

        let is_valid = value.is_instance_of::<PyDate>()
            || value.is_instance_of::<PyTime>()
            || value.is_instance_of::<PyDateTime>();

        if !is_valid {
            return Ok(Some("".as_content()));
        }

        let date_format_fn = DATE_FORMAT.import(py, "django.utils.dateformat", "format")?;

        let formatted = match date_format_fn.call1((value, fmt)) {
```

tests/filters/test_date.py
```diff
@@ -0,0 +1,194 @@
...
+    ],
+)
+def test_date_functional_scenarios(assert_render, template, context, expected):
+    assert_render(template, context, expected)
+
+
+@pytest.mark.parametrize(
+    "template,context,expected",
+    [
+        pytest.param("{{ d|date:'Y' }}", {"d": 123}, "", id="test_date_int_input"),
```
LilyFirefly: We can also test with a float and with literals in the template:
```suggestion
        pytest.param("{{ d|date:'Y' }}", {"d": 123}, "", id="test_date_int_input"),
        pytest.param("{{ d|date:'Y' }}", {"d": 1.2}, "", id="test_date_float_input"),
        pytest.param("{{ 123|date:'Y' }}", {}, "", id="test_date_int_literal"),
        pytest.param("{{ 1.2|date:'Y' }}", {}, "", id="test_date_float_literal"),
```

src/render/filters.rs
```diff
@@ -274,6 +278,55 @@ impl ResolveFilter for DefaultIfNoneFilter {
...
+            Ok(res) => res,
+            Err(e) => {
+                if e.is_instance_of::<pyo3::exceptions::PyAttributeError>(py) {
+                    return Ok(Some("".as_content()));
+                }
+                return Err(PyRenderError::PyErr(e));
```
LilyFirefly: I'd like us to use `annotate` to add miette formatting to the error for better debugging:

```suggestion
                return Err(PyRenderError::PyErr(e.annotate(py, self.at, "here", template)));
```

`DateFilter` will need an `at` attribute. See `DefaultFilter` for an example.

LilyFirefly (commented):
This is almost there! Just a few suggestions and it'll be ready to merge!

p-r-a-v-i-n:
> Thanks for adding these filters! If you think you're ready for something a bit harder, I think you could take a look at adding a tag from #2.

Yeah sure, I would love to explore that too. It seems difficult i have gone through load tag recently. I will try slowly slowly. may be with lorem

PR #274, opened 2025-12-25
chore(django-dtl): move the empty tag case into django-dtl

christophehenry:
This continues what I started in #272. This simply move the empty tag case in `django-dtl`.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -35,7 +35,6 @@ use crate::filters::UpperFilter;
 use crate::filters::WordcountFilter;
 use crate::filters::WordwrapFilter;
 use crate::filters::YesnoFilter;
-use dtl_lexer::START_TAG_LEN;
 use dtl_lexer::common::{LexerError, text_content_at, translated_text_content_at};
 use dtl_lexer::core::{Lexer, TokenType};
 use dtl_lexer::tag::autoescape::{AutoescapeEnabled, AutoescapeError, lex_autoescape_argument};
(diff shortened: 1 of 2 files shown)


UnknownPlatypus:
Generated parser are usually less performant than hand-written parser and harder to optimize. Given this project goal is to implement a more performant django template parser and that a bunch of hand-written parsing logic is already implemented, I'm not sure it's a good fit.

[Ruff switched from a generated parser to a hand-written recursive descent parser](https://astral.sh/blog/ruff-v0.4.0) and got a 2x perf improvement. The whole blog post is interesting, speed was not the only argument to switch

src/parse.rs
```diff
@@ -1235,61 +1236,53 @@ impl<'t, 'py> Parser<'t, 'py> {
...
-            }
-        };
-        let Some((tag, parts)) = maybe_tag else {
-            return Err(ParseError::EmptyTag { at: at.into() }.into());
-        };
-        Ok(match self.template.content(tag.at) {
-            "url" => Either::Left(self.parse_url(at, parts)?),
-            "load" => Either::Left(self.parse_load(at, parts)?),
-            "autoescape" => Either::Left(self.parse_autoescape(at, parts)?),
+        let tag = lex_tag(tag, at.0 + START_TAG_LEN)?;
```
LilyFirefly: Instead of `impl From<TagLexerError> for PyParseError` we could do:

```suggestion
        let tag = lex_tag(tag, at.0 + START_TAG_LEN).map_err(ParseError::from)?;
```

christophehenry: Ah yes. It's more elegant indeed. I edited my commit to include that.

PR #275, opened 2025-12-25
Add inline snapshot

UnknownPlatypus:
Fixes https://github.com/LilyFirefly/django-rusty-templates/pull/268#issuecomment-3689999948

This is even simpler than I thought:
- running `pytest` with empty `snapshot()` will automatically populate them (and fail the test to notify that)
- running `pytest` with the content of a snapshot changing will prompt you to validate the update (a la `cargo insta review`)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dev = [
     "pytest-cov",
     "ruff",
     "pre-commit",
-    "inline-snapshot"
+    "inline-snapshot",
 ]
 
 [tool.ruff]


pyproject.toml
```diff
@@ -29,7 +29,8 @@ dev = [
     "pytest-django",
     "pytest-cov",
     "ruff",
-    "pre-commit"
+    "pre-commit",
+    "inline-snapshot"
```
LilyFirefly: Can we have a trailing comma in toml, to avoid messy diffs?

```suggestion
    "inline-snapshot",
```

PR #276, opened 2025-12-26
Implemented `{% lorem %}` tag

p-r-a-v-i-n:
part of #2

### DIFF
--- a/tests/tags/test_lorem.py
+++ b/tests/tags/test_lorem.py
@@ -1,5 +1,6 @@
 import pytest
 from django.utils.lorem_ipsum import COMMON_P, WORDS, COMMON_WORDS
+from inline_snapshot import snapshot
 
 
 def test_lorem_words(assert_render):
@@ -445,3 +446,24 @@ def test_lorem_words_with_negative_count(render_output):
     template = "{% lorem -5 w %}"
     output = render_output(template=template, context={})
     assert output == " ".join(COMMON_WORDS[:-5])
+
+
+def test_lorem_render_error(assert_render_error):
+    template = "{% lorem count %}"
+    context = {"count": lambda: 1 / 0}
+    django_message = snapshot("division by zero")
+    rusty_message = snapshot("""\
+  × division by zero
+   ╭────
+ 1 │ {% lorem count %}
+   ·          ──┬──
+   ·            ╰── here
+   ╰────
+""")
+    assert_render_error(
+        template=template,
+        context=context,
+        exception=ZeroDivisionError,
+        django_message=django_message,
+        rusty_message=rusty_message,
+    )


tests/tags/test_lorem.py
```diff
@@ -0,0 +1,166 @@
...
+
+    django_message = "Incorrect format for 'lorem' tag"
+
+    rusty_message = """\
+  × Incorrect format for 'lorem' tag
+   ╭────
+ 1 │ {% lorem 1 2 3 4 %}
+   ·            ┬
+   ·            ╰── here
+   ╰────
```
LilyFirefly: I think we can provide better error messages than Django. For example, I think this can be something like:

```suggestion
  × Incorrect format for 'lorem' tag: 'count' argument was provided more than once
   ╭────
 1 │ {% lorem 1 2 3 4 %}
   ·          ┬ ┬
   ·          │ ╰── second 'count'
   ·          ╰── first 'count'
   ╰────
  help: Try removing the second 'count'
```

tests/tags/test_lorem.py
```diff
@@ -0,0 +1,166 @@
...
+def test_lorem_duplicate_method(assert_parse_error):
+    template = "{% lorem 2 w p %}"
+    django_message = "Incorrect format for 'lorem' tag"
+    rusty_message = """\
+  × Incorrect format for 'lorem' tag
+   ╭────
+ 1 │ {% lorem 2 w p %}
+   ·              ┬
+   ·              ╰── here
+   ╰────
```
LilyFirefly:
```suggestion
  × Incorrect format for 'lorem' tag: 'method' argument was provided more than once
   ╭────
 1 │ {% lorem 2 w p %}
   ·            ┬ ┬
   ·            │ ╰── second 'method'
   ·            ╰── first 'method'
   ╰────
  help: Try removing the second 'method'
```

tests/tags/test_lorem.py
```diff
@@ -0,0 +1,166 @@
...
+def test_lorem_duplicate_random(assert_parse_error):
+    template = "{% lorem 2 p random random %}"
+    django_message = "Incorrect format for 'lorem' tag"
+    rusty_message = """\
+  × Incorrect format for 'lorem' tag
+   ╭────
+ 1 │ {% lorem 2 p random random %}
+   ·                     ───┬──
+   ·                        ╰── here
+   ╰────
```
LilyFirefly:
```suggestion
  × Incorrect format for 'lorem' tag: 'random' was provided more than once
   ╭────
 1 │ {% lorem 2 p random random %}
   ·              ───┬── ───┬──
   ·                 │      ╰── here
   ·                 ╰── here
   ╰────
  help: Try removing the second 'random'
```

tests/tags/test_lorem.py
LilyFirefly, line 1: We should also test `{% lorem w p %}` and `{% lorem random random %}` - in these I think the first entry should be treated as a variable (looked up from `context`) and the second entry as `method` or `random` respectively.

We'll also need a test for a filter on `count`. E.g: `{% lorem count|default:3 %}`

tests/tags/test_lorem.py
```diff
@@ -0,0 +1,166 @@
+from django.utils.lorem_ipsum import COMMON_P, WORDS
+
+
+def test_lorem_words(render_output):
+    template = "{% lorem 3 w %}"
+    output = render_output(
+        template=template,
+        context={},
+    )
+    assert "lorem ipsum dolor" == output
```
LilyFirefly: It doesn't really matter, but it's more consistent to use `assert_render` where possible:

```suggestion
def test_lorem_words(assert_render):
    template = "{% lorem 3 w %}"
    assert_render(
        template=template,
        context={},
        expected="lorem ipsum dolor",
    )
```

Some of the other tests below can also use `assert_render`.

src/utils.rs
```diff
@@ -19,3 +21,292 @@ impl<T> PyResultMethods<T> for PyResult<T> {
...
+
+        while remaining > 0 {
+            let take = remaining.min(WORDS.len());
+            let sampled = WORDS.choose_multiple(&mut rng, take);
+            word_list.extend(sampled.cloned());
+            remaining -= take;
+        }
+    }
+
+    word_list.join(" ")
```
LilyFirefly: I don't think we need to create a `Vec` from `COMMON_WORDS` if `count` is less than `COMMON_WORDS.len()`:

```suggestion
pub fn words(count: usize, common: bool) -> String {
    if common && count <= COMMON_WORDS.len() {
        return COMMON_WORDS[..count].join(" ");
    }

    let mut rng = rand::thread_rng();
    let mut word_list: Vec<&str> = if common {
        COMMON_WORDS.to_vec()
    } else {
        Vec::new()
    };

    let mut remaining = count - word_list.len();

    while remaining > 0 {
        let take = remaining.min(WORDS.len());
        let sampled = WORDS.choose_multiple(&mut rng, take);
        word_list.extend(sampled.cloned());
        remaining -= take;
    }

    word_list.join(" ")
```

src/utils.rs
```diff
@@ -19,3 +21,292 @@ impl<T> PyResultMethods<T> for PyResult<T> {
...
+}
+
+pub fn words(count: usize, common: bool) -> String {
+    let mut rng = rand::thread_rng();
+
+    let mut word_list: Vec<&str> = if common {
+        COMMON_WORDS.to_vec()
+    } else {
+        Vec::new()
+    };
```
LilyFirefly: We can use [`Vec::with_capacity`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity):

```suggestion
    let mut word_list: Vec<&str> = Vec::with_capacity(count);
    if common {
        word_list.extend(&COMMON_WORDS);
    }
```

src/utils.rs
```diff
@@ -19,3 +21,292 @@ impl<T> PyResultMethods<T> for PyResult<T> {
...
+pub fn paragraph() -> String {
+    let num_sentences = rand::thread_rng().gen_range(1..=4);
+    (0..num_sentences)
+        .map(|_| sentence())
+        .collect::<Vec<String>>()
+        .join(" ")
+}
+
+pub fn paragraphs(count: usize, common: bool) -> Vec<String> {
+    let mut paras = Vec::new();
```
LilyFirefly:
```suggestion
    let mut paras = Vec::with_capacity(count);
```

src/utils.rs
```diff
@@ -19,3 +21,292 @@ impl<T> PyResultMethods<T> for PyResult<T> {
...
+    "et",
+    "dolore",
+    "magna",
+    "aliqua",
+];
+
+pub fn sentence() -> String {
+    let mut rng = rand::thread_rng();
+    let num_sections = rng.gen_range(1..=5);
+    let mut sections = Vec::new();
```
LilyFirefly:
```suggestion
    let mut sections = Vec::with_capacity(num_sections);
```

src/utils.rs
```diff
@@ -19,3 +21,292 @@ impl<T> PyResultMethods<T> for PyResult<T> {
...
+            .cloned()
+            .collect();
+        sections.push(selected_words.join(" "));
+    }
+
+    let sentence = sections.join(", ");
+    let first_char = sentence
+        .chars()
+        .next()
+        .unwrap()
```
LilyFirefly: We should use `expect` here instead of `unwrap`.

src/utils.rs
```diff
@@ -19,3 +21,292 @@ impl<T> PyResultMethods<T> for PyResult<T> {
...
+        .to_uppercase()
+        .collect::<String>();
```
LilyFirefly: We don't need to convert the `char` back to a `String` here, since the `format!` macro can handle `char`:

```suggestion
        .to_uppercase();
```

src/utils.rs
```diff
@@ -19,3 +21,292 @@ impl<T> PyResultMethods<T> for PyResult<T> {
...
+    let rest = &sentence[1..];
```
LilyFirefly: It's technically more correct to use the length of `first_char`:

```suggestion
    let rest = &sentence[first_char.len_utf8()..];
```

Otherwise, this can cause problems with non-ascii characters. In this case I don't think it's actually a problem, since `WORDS` are all ascii, but I've been caught out with this mistake elsewhere when it did matter.

src/utils.rs
```diff
@@ -19,3 +21,292 @@ impl<T> PyResultMethods<T> for PyResult<T> {
...
+    let punctuation = if rng.gen_bool(0.5) { "?" } else { "." };
+
+    format!("{}{}{}", first_char, rest, punctuation)
```
LilyFirefly:
```suggestion
    format!("{first_char}{rest}{punctuation}")
```

src/utils.rs
LilyFirefly, line 1: I'd put the `lorem` code in its own file, say `src/lorem.rs`.

src/parse.rs
```diff
@@ -1287,6 +1309,7 @@ impl<'t, 'py> Parser<'t, 'py> {
                 parts,
             }),
             "include" => Either::Left(self.parse_include(at, parts)?),
+            "lorem" => Either::Left(TokenTree::Tag(Tag::Lorem(lex_lorem(self.template, parts)?))),
```
LilyFirefly: We'll need a parsing step in addition to the lexing step. For example, `count` can be a variable or an integer and we'll need to handle that here.

LilyFirefly (commented):
This is an excellent start! I've spotted a few things we still need to handle and I have some suggestions for improvements:

p-r-a-v-i-n:
Hi @LilyFirefl, Till next Sunday I will be busy with my family so  I would need some time to complete this work. I know you wouldn't mind but still I wanna let you know that bcs I'm really enjoying working here . I will be doing small small changes as time permits. I have updated the small easy changes.

dtl-lexer/src/lorem.rs
LilyFirefly, line 1: None of this is lexing code (aka tokenizing), so it shouldn't be in the `dtl-lexer` crate. I'd put it in `src/lorem.rs` (instead of `dtl-lexer/src/lorem.rs`) or in `src/render/lorem.rs`.

Also, please move code between files in a separate commit to refactoring the code, it makes it easier to review commit-by-commit.

dtl-lexer/src/lorem.rs
```diff
@@ -0,0 +1,288 @@
...
+            .copied()
+            .collect();
+
+        sections.push(selected_words.join(" "));
+    }
+
+    let mut sentence = sections.join(", ");
+
+    if let Some(first) = sentence.chars().next() {
+        let upper = first.to_uppercase().to_string();
```
LilyFirefly: We don't need to convert to a `String` here:

```suggestion
        let upper = first.to_uppercase();
```

dtl-lexer/src/lorem.rs
```diff
@@ -0,0 +1,288 @@
...
+}
+
+pub fn paragraphs(count: usize, common: bool) -> Vec<String> {
+    let mut paras = Vec::with_capacity(count);
+
+    for i in 0..count {
+        if common && i == 0 {
+            paras.push(COMMON_P.to_string());
+        } else {
+            paras.push(paragraph());
```
LilyFirefly: We can use [`Cow`](https://doc.rust-lang.org/std/borrow/enum.Cow.html) here to avoid the `to_string()` call:

```suggestion
pub fn paragraphs(count: usize, common: bool) -> Vec<Cow<'static, str>> {
    let mut paras = Vec::with_capacity(count);

    for i in 0..count {
        if common && i == 0 {
            paras.push(Cow::Borrowed(COMMON_P));
        } else {
            paras.push(Cow::Owned(paragraph()));
```

dtl-lexer/src/lorem.rs
```diff
@@ -0,0 +1,288 @@
...
+pub fn words(count: usize, common: bool) -> String {
+    if common && count <= COMMON_WORDS.len() {
+        return COMMON_WORDS[..count].join(" ");
+    }
+    let mut rng = rand::thread_rng();
+    let mut word_list: Vec<&str> = Vec::with_capacity(count);
+    if common {
+        word_list.extend(&COMMON_WORDS);
+    }
+    let mut remaining = count - word_list.len();
```
LilyFirefly: I think this is a bit clearer to read:

```suggestion
    let mut remaining = count;
    if common {
        word_list.extend(&COMMON_WORDS);
        remaining -= word_list.len();
    }
```

dtl-lexer/src/lorem.rs
```diff
@@ -0,0 +1,288 @@
...
+    while remaining > 0 {
+        let take = remaining.min(WORDS.len());
+        let sampled = WORDS.choose_multiple(&mut rng, take);
+        word_list.extend(sampled.cloned());
+        remaining -= take;
```
LilyFirefly: We could combine `remaining` and `count`:

```suggestion
pub fn words(mut count: usize, common: bool) -> String {
    if common && count <= COMMON_WORDS.len() {
        return COMMON_WORDS[..count].join(" ");
    }
    let mut rng = rand::thread_rng();
    let mut word_list: Vec<&str> = Vec::with_capacity(count);
    if common {
        word_list.extend(&COMMON_WORDS);
        count -= word_list.len();
    }
    while count > 0 {
        let take = count.min(WORDS.len());
        let sampled = WORDS.choose_multiple(&mut rng, take);
        word_list.extend(sampled.copied());
        count -= take;
```

dtl-lexer/src/lorem.rs
```diff
@@ -0,0 +1,288 @@
...
+    let mut rng = rand::thread_rng();
+    let mut word_list: Vec<&str> = Vec::with_capacity(count);
+    if common {
+        word_list.extend(&COMMON_WORDS);
+    }
+    let mut remaining = count - word_list.len();
+    while remaining > 0 {
+        let take = remaining.min(WORDS.len());
+        let sampled = WORDS.choose_multiple(&mut rng, take);
+        word_list.extend(sampled.cloned());
```
LilyFirefly: I think it's a bit clearer that this is really cheap if we use `copied` instead of `cloned`. It's also more consistent with `sentence` where you used `copied`.

```suggestion
        word_list.extend(sampled.copied());
```

dtl-lexer/src/tag/lorem.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum LoremMethod {
+    Words,
+    Paragraphs,
+    Blocks,
+}
+
+#[derive(Debug, Error, PartialEq, Eq)]
```
LilyFirefly: We should `derive(Diagnostic)` (from `miette`) too.

dtl-lexer/src/tag/lorem.rs
```diff
@@ -0,0 +1,134 @@
...
+    InvalidFormat { at: At },
+
+    #[error("Incorrect format for 'lorem' tag: 'random' was provided more than once")]
+    DuplicateRandom { first: At, second: At },
+
+    #[error("Incorrect format for 'lorem' tag: 'method' argument was provided more than once")]
+    DuplicateMethod { first: At, second: At },
+
+    #[error("Incorrect format for 'lorem' tag: 'count' argument was provided more than once")]
+    DuplicateCount { first: At, second: At },
```
LilyFirefly: These should use `miette`'s `SourceSpan` instead of `At`.

src/parse.rs
```diff
@@ -943,6 +952,38 @@ pub enum ParseError {
...
+    },
+
+    #[error("Incorrect format for 'lorem' tag: 'count' argument was provided more than once")]
+    #[diagnostic(help("Try removing the second 'count'"))]
+    LoremDuplicateCount {
+        #[label("first 'count'")]
+        first_at: SourceSpan,
+        #[label("second 'count'")]
+        second_at: SourceSpan,
+    },
```
LilyFirefly: We should be able to replace most/all of these with:

```rust
    #[error(transparent)]
    #[diagnostic(transparent)]
    LoremError(#[from] LoremError),
```

src/parse.rs
```diff
@@ -953,6 +994,38 @@ pub enum PyParseError {
...
+            },
+
+            LoremError::DuplicateCount { first, second } => ParseError::LoremDuplicateCount {
+                first_at: first.into(),
+                second_at: second.into(),
+            },
+        }
+    }
+}
+
```
LilyFirefly: And this should be unnecessary.

src/parse.rs
```diff
@@ -1229,6 +1302,23 @@ impl<'t, 'py> Parser<'t, 'py> {
         }
         Ok(var)
     }
+    fn parse_lorem(&mut self, _at: At, parts: TagParts) -> Result<Lorem, PyParseError> {
+        let token = lex_lorem(self.template, parts)?;
```
LilyFirefly: This might need to be:

```suggestion
        let token = lex_lorem(self.template, parts).map_err(ParseError::from)?;
```

tests/tags/test_lorem.py
```diff
@@ -0,0 +1,206 @@
...
+    assert COMMON_P not in output
+    assert output.count("<p>") == 2
+
+
+def test_lorem_zero_count(assert_render):
+    template = "{% lorem 0 w %}"
+    assert_render(template=template, context={}, expected="")
+
+
+66
```
LilyFirefly:
```suggestion
```

tests/tags/test_lorem.py
```diff
@@ -0,0 +1,206 @@
...
+def test_lorem_method_before_count(assert_parse_error):
+    template = "{% lorem p 2 %}"
+    django_message = "Incorrect format for 'lorem' tag"
+    rusty_message = """\
+  × Incorrect format for 'lorem' tag
+   ╭────
+ 1 │ {% lorem p 2 %}
+   ·            ┬
+   ·            ╰── here
+   ╰────
```
LilyFirefly: We should have a better error message here too. Either something explaining that `count` should come before `method` or something saying `count` is specified twice.

tests/tags/test_lorem.py
```diff
@@ -0,0 +1,206 @@
...
+def test_lorem_random_before_count(assert_parse_error):
+    template = "{% lorem random 2 %}"
+    django_message = "Incorrect format for 'lorem' tag"
+    rusty_message = """\
+  × Incorrect format for 'lorem' tag
+   ╭────
+ 1 │ {% lorem random 2 %}
+   ·                 ┬
+   ·                 ╰── here
+   ╰────
```
LilyFirefly: We should have a better error message here too.

tests/tags/test_lorem.py
```diff
@@ -0,0 +1,206 @@
...
+    assert len(output.split(" ")) == 2
+
+
+def test_lorem_with_filter(render_output):
+    template = "{% lorem count|default:3 w %}"
+    output_default = render_output(template=template, context={})
+    assert len(output_default.split(" ")) == 3
+
+    output_provided = render_output(template=template, context={"count": 1})
+    assert output_provided == "lorem"
```
LilyFirefly: This should be two tests (or maybe one test with clever use of `parametrize`).

src/parse.rs
```diff
@@ -1229,6 +1302,23 @@ impl<'t, 'py> Parser<'t, 'py> {
...
+
+        let count = if let Some(count_at) = token.count_at {
+            let content = self.template.content(count_at);
+            self.parse_variable(content, count_at, count_at.0)?
```
LilyFirefly: It would be good to have a test for the error path here.

src/render/tags.rs
```diff
@@ -665,6 +667,34 @@ impl Render for Tag {
             Self::SimpleTag(simple_tag) => simple_tag.render(py, template, context)?,
             Self::SimpleBlockTag(simple_tag) => simple_tag.render(py, template, context)?,
             Self::Url(url) => url.render(py, template, context)?,
+            Self::Lorem(lorem) => {
+                let count_content = lorem.count.resolve(
+                    py,
+                    template,
+                    context,
+                    ResolveFailures::IgnoreVariableDoesNotExist,
+                )?;
```
LilyFirefly, line 676: It would be good to test this error path.

dtl-lexer/src/lorem.rs
```diff
@@ -0,0 +1,288 @@
...
+
+pub fn words(count: usize, common: bool) -> String {
+    if common && count <= COMMON_WORDS.len() {
+        return COMMON_WORDS[..count].join(" ");
+    }
+    let mut rng = rand::thread_rng();
+    let mut word_list: Vec<&str> = Vec::with_capacity(count);
+    if common {
+        word_list.extend(&COMMON_WORDS);
+    }
```
LilyFirefly, line 282: We should have a test that hits this code.

dtl-lexer/src/lorem.rs
```diff
@@ -0,0 +1,288 @@
...
+        sections.push(selected_words.join(" "));
+    }
+
+    let mut sentence = sections.join(", ");
+
+    if let Some(first) = sentence.chars().next() {
+        let upper = first.to_uppercase().to_string();
+        let rest = &sentence[first.len_utf8()..];
+        sentence = format!("{upper}{rest}");
+    }
```
LilyFirefly: We know this can't fail:

```suggestion
    let first = sentence.chars().next().expect("A sentence should have at least one character");
    let upper = first.to_uppercase();
    let rest = &sentence[first.len_utf8()..];
    sentence = format!("{upper}{rest}");
```

tests/tags/test_lorem.py
```diff
@@ -0,0 +1,329 @@
...
+def test_lorem_string_garbage_defaults_to_one(render_output):
+    """A string that isn't a number should default to 1."""
+    template = "{% lorem count w %}"
+    output = render_output(template=template, context={"count": "abc"})
+    assert output == "lorem"
+
+
+# def test_lorem_extreme_count_defaults_to_one(render_output):
+#     # A number way larger than a 64-bit integer
+#     template = "{% lorem 99999999999999999999999999999999999999 w %}"
```
p-r-a-v-i-n, line 412: well django doesn't handled this yet i guess , so i commented out here. my system hangs

src/render/lorem.rs
```diff
@@ -0,0 +1,455 @@
...
+
+static WORDS: [&str; 182] = [
+    "exercitationem",
+    "perferendis",
+    "perspiciatis",
+    "laborum",
+    "eveniet",
+    "sunt",
+    "iure",
+    "name",
```
LilyFirefly:
```suggestion
    "nam",
```

src/render/lorem.rs
```diff
@@ -0,0 +1,455 @@
...
+    "magnam",
+    "recusandae",
+    "quod",
+    "explicabo",
+    "doloremque",
+    "aperiam",
+    "consequatur",
+    "asperiores",
+    "commodi",
+    "option",
```
LilyFirefly:
```suggestion
    "optio",
```

LilyFirefly (commented):
I think these typos (autocorrect?) are why there's some random test failures.

src/parse.rs
```diff
@@ -1229,6 +1248,30 @@ impl<'t, 'py> Parser<'t, 'py> {
         }
         Ok(var)
     }
+    fn parse_lorem(&mut self, _at: At, parts: TagParts) -> Result<Lorem, PyParseError> {
```
LilyFirefly, line 1251: We should have a blank line between methods:

```suggestion

    fn parse_lorem(&mut self, _at: At, parts: TagParts) -> Result<Lorem, PyParseError> {
```

src/parse.rs
```diff
@@ -1229,6 +1248,30 @@ impl<'t, 'py> Parser<'t, 'py> {
...
+
+        let count = if let Some(count_at) = token.count_at {
+            let content = self.template.content(count_at);
+            if content.ends_with('|') {
+                return Err(PyParseError::from(ParseError::LoremError(
+                    LoremError::InvalidRemainder {
+                        _at: (count_at.0 + content.len() - 1, 1).into(),
+                    },
+                )));
+            }
```
LilyFirefly: This will become unnecessary with #280, which handles this case in `lex_variable_or_filter` (which is used in `self.parse_variable` below).

tests/tags/test_lorem.py
```diff
@@ -0,0 +1,341 @@
...
+def test_lorem_default_negative_count_defaults_to_one(render_output):
+    template = "{% lorem -5 %}"
+    output = render_output(template=template, context={})
+    assert output == ""
+
+
+def test_lorem_word_negative_count_defaults_to_one(render_output):
+    template = "{% lorem -5 w %}"
+    output = render_output(template=template, context={})
+    assert output == " ".join(COMMON_WORDS[:-5])
```
LilyFirefly: This test name doesn't seem right for the implementation.

Also, interesting edge case! Django doesn't document that it ever treats negative numbers specially.

tests/tags/test_lorem.py
```diff
@@ -0,0 +1,341 @@
...
+
+    assert len(words_list) == 20
+
+    assert words_list[:19] == list(COMMON_WORDS)
+
+
+def test_lorem_default_negative_count_defaults_to_one(render_output):
+    template = "{% lorem -5 %}"
+    output = render_output(template=template, context={})
+    assert output == ""
```
LilyFirefly: This test name doesn't seem right for the implementation.

tests/tags/test_lorem.py
LilyFirefly: I think it's worth reporting this as a bug upstream - a system hang for a too-large count here isn't great.

tests/tags/test_lorem.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+def test_lorem_with_complex_filter_parsing(render_output):
+    template = "{% lorem val|add:2|default:5 w %}"
+
+    output_a = render_output(template=template, context={"val": 1})
+    assert len(output_a.split(" ")) == 3
+
+    output_b = render_output(template=template, context={})
+    assert len(output_b.split(" ")) == 5
```
LilyFirefly: This can use `parametrize`, like `test_lorem_with_filter` does.

src/render/tags.rs
```diff
@@ -665,6 +666,49 @@ impl Render for Tag {
...
+                let val = count_content
+                    .and_then(|c| c.to_bigint())
+                    .and_then(|n| n.to_i64())
+                    .unwrap_or(1);
+
+                let text = match lorem.method {
+                    LoremMethod::Words => {
+                        let final_count = if val < 0 {
+                            // common words len is 19
+                            (19 + val).max(0) as usize
```
LilyFirefly: It's much better to calculate this here than use a hardcoded magic number. `.len()` is `const`, so this will get optimised away at compile time anway.

```suggestion
                            (COMMON_WORDS.len() + val).max(0) as usize
```

src/render/tags.rs
```diff
@@ -665,6 +666,49 @@ impl Render for Tag {
...
+                        } else {
+                            val as usize
+                        };
+                        words(final_count, lorem.common)
+                    }
+                    LoremMethod::Paragraphs | LoremMethod::Blocks => {
+                        if val <= 0 {
+                            String::new()
```
LilyFirefly: We don't need to allocate a new `String` here:

```suggestion
                            return Cow::Borrowed("");
```

tests/tags/test_lorem.py
```diff
@@ -0,0 +1,341 @@
...
+    template = "{% lorem p 2 %}"
+    django_message = "Incorrect format for 'lorem' tag"
+    rusty_message = """\
+  × Incorrect format for 'lorem' tag: count must come before method or random
+   ╭────
+ 1 │ {% lorem p 2 %}
+   ·            ┬
+   ·            ╰── count must come first
+   ╰────
+  help: Move the count argument before the method
```
LilyFirefly: I think the error message would be better like this:

```suggestion
  × Incorrect format for 'lorem' tag: 'count' must come before the 'method' argument
   ╭────
 1 │ {% lorem p 2 %}
   ·          ┬ ┬
   ·          │ ╰── count
   ·          ╰── method
   ╰────
  help: Move the 'count' argument before the 'method' argument
```

tests/tags/test_lorem.py
```diff
@@ -0,0 +1,341 @@
...
+    template = "{% lorem random 2 %}"
+    django_message = "Incorrect format for 'lorem' tag"
+    rusty_message = """\
+  × Incorrect format for 'lorem' tag: count must come before method or random
+   ╭────
+ 1 │ {% lorem random 2 %}
+   ·                 ┬
+   ·                 ╰── count must come first
+   ╰────
+  help: Move the count argument before the method
```
LilyFirefly:
```suggestion
  × Incorrect format for 'lorem' tag: 'count' must come before the 'random' argument
   ╭────
 1 │ {% lorem random 2 %}
   ·          ───┬── ┬
   ·             │   ╰── count
   ·             ╰── random
   ╰────
  help: Move the 'count' argument before the 'random' argument
```

src/render/lorem.rs
```diff
@@ -0,0 +1,455 @@
...
+        }
+    }
+
+    Ok(LoremToken {
+        at: parts.at,
+        count_at,
+        method,
+        common,
+    })
+}
```
LilyFirefly: I think I was unclear before. This part of this file *should* be in the `dtl-lexer` crate. It's the implementation of `words` and `paragraphs` that should be here in `src/render/lorem.rs`.

src/render/lorem.rs
```diff
@@ -0,0 +1,455 @@
...
+        _second: SourceSpan,
+    },
+    #[error("Invalid filter: ''")]
+    InvalidRemainder {
+        #[label("expected filter name after '|'")]
+        _at: SourceSpan,
+    },
+}
+
+pub fn lex_lorem(template: TemplateString<'_>, parts: TagParts) -> Result<LoremToken, LoremError> {
```
LilyFirefly: I think it might be valuable to refactor to create a `Lexer` struct that has an `impl Iterator` which yields `LoremToken`s. The `LoremToken` could look like:

```rust
pub enum LoremTokenType {
    Count,
    Method(LoremMethod),
    Random,
}

pub struct LoremToken {
    at: At,
    type: LoremTokenType,
}
```

The logic to turn a stream of tokens into a `Lorem` instance would then move to `parse_lorem` in `parse.rs`. The error handling would be split between the lexer and the parser, depending on where it's easiest to handle.

As part of this, I'd return a `Method` or `Random` token from the `LoremLexer` even if the parser then decides its actually representing a `Count`.

LilyFirefly (commented):
I know I keep coming back with a lot of feedback, but we're really getting there! Great work!

tests/tags/test_lorem.py
p-r-a-v-i-n: what's your take on this, how do you see this ?

src/render/lorem.rs
p-r-a-v-i-n: I may sound dump here asking again , should i move this part in `dtl-lexer/src/tag/lorem.rs` ? and rest remain in `src/render/lorem.rs`

tests/tags/test_lorem.py
UnknownPlatypus: Yes, it could be abused if user controlled (seems unlikely but still)

src/render/lorem.rs
LilyFirefly: Yes, that's right!

tests/tags/test_lorem.py
LilyFirefly: I think we should match Django's behaviour here, as you've done. It's probably unintentional behaviour in Django itself, but I think it's probably not worth changing.

src/render/lorem.rs
p-r-a-v-i-n: pre-commit bot will autocorrect this again

LilyFirefly (approved):
Thanks again for this!

I decided to refactor the lexer to avoid needing the heap allocation of `Vec`, which led to me finding a few more edge-cases, so I've added some more tests too.

PR #278, opened 2025-12-30
Add now filter

p-r-a-v-i-n:
Part of #2

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1389,6 +1389,10 @@ impl<'t, 'py> Parser<'t, 'py> {
         let asvar = lexer.lex_variable().map_err(ParseError::from)?;
         lexer.extra_token().map_err(ParseError::from)?;
         let raw = self.template.content(format_at);
+        // Django always trims the first and last character without further
+        // validation, so so do we. This will become unnecessary if Django
+        // starts supporting variables in the now tag.
+        // https://github.com/django/new-features/issues/115
         let format = if raw.len() >= 2 {
             &raw[1..raw.len() - 1]
         } else {
--- a/tests/tags/test_now.py
+++ b/tests/tags/test_now.py
@@ -483,6 +483,16 @@ def test_now_invalid_variable_name(assert_render):
 
 
 @time_machine.travel(FIXED_TIME)
+def test_now_invalid_variable_name_error(assert_render):
+    template = "{% now invalid-name %}"
+    assert_render(
+        template=template,
+        context={},
+        expected=r"1vp.m.Thursday0008-1p.m.01",
+    )
+
+
+@time_machine.travel(FIXED_TIME)
 def test_now_incomplete_translated_string(template_engine, assert_render):
     template = '{% now _("j n Y %}'
     if template_engine.name != "django":


dtl-lexer/src/tag/now.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub at: At,
+    pub token_type: NowTokenType,
+}
+
+pub struct NowLexer<'t> {
+    rest: &'t str,
+    byte: usize,
+    seen_format: bool,
+    seen_as: bool,
+    seen_variable: bool,
```
LilyFirefly: I think it would make sense to merge these into a single `next_token_type`:

```suggestion
    next_token_type: Option<NowTokenType>,
```

dtl-lexer/src/tag/now.rs
```diff
@@ -0,0 +1,107 @@
...
+impl<'t> NowLexer<'t> {
+    pub fn new(template: TemplateString<'t>, parts: TagParts) -> Self {
+        let content = template.content(parts.at);
+        let skip = content.next_non_whitespace();
+        Self {
+            rest: &content[skip..],
+            byte: parts.at.0 + skip,
+            seen_format: false,
+            seen_as: false,
+            seen_variable: false,
```
LilyFirefly:
```suggestion
            next_token_type: NowTokenType::Format,
```

dtl-lexer/src/tag/now.rs
```diff
@@ -0,0 +1,107 @@
...
+            Err(NowError::Syntax { at: at.into() })
+        };
+
+        let token_type = match token_type {
+            Ok(t) => t,
+            Err(e) => {
+                self.rest = "";
+                return Some(Err(e));
+            }
+        };
```
LilyFirefly:
```suggestion
        let token_type = match self.next_token_type {
            Some(token_type) => token_type,
            None => {
                self.rest = "";
                return Some(Err(NowError::Syntax { at: at.into() }))
            }
        };
        self.next_token_type = token_type.next();
```

dtl-lexer/src/tag/now.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub asvar: Option<At>,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum NowTokenType {
+    Format,
+    As,
+    Variable,
+}
+
```
LilyFirefly:
```suggestion
pub enum NowTokenType {
    Format,
    As,
    Variable,
}

impl NowTokenType {
    fn next(&self) -> Option<Self> {
        match self {
            Self::Format => Some(Self::As),
            Self::As => Some(Self::Variable),
            Self::Variable => None,
        }
    }
}
```

dtl-lexer/src/tag/now.rs
LilyFirefly: We should also ensure that `content` is exactly the string `"as"` when `token_type` is `NowTokenType::As` and return an error otherwise. The same goes for returning an error when `content` has an invalid form when the `token_type` is `NowTokenType::Variable`.

dtl-lexer/src/tag/now.rs
```diff
@@ -0,0 +1,107 @@
...
+    seen_variable: bool,
+}
+
+impl<'t> NowLexer<'t> {
+    pub fn new(template: TemplateString<'t>, parts: TagParts) -> Self {
+        let content = template.content(parts.at);
+        let skip = content.next_non_whitespace();
+        Self {
+            rest: &content[skip..],
+            byte: parts.at.0 + skip,
```
LilyFirefly: `TagParts.at` is constructed so that it doesn't start with whitespace (see https://github.com/LilyFirefly/django-rusty-templates/blob/main/dtl-lexer/src/tag.rs#L53-L87), so this can be simplified:

```suggestion
        Self {
            rest: template.content(parts.at),
            byte: parts.at.0,
```

src/parse.rs
```diff
@@ -1329,6 +1335,52 @@ impl<'t, 'py> Parser<'t, 'py> {
...
+    fn parse_now(&mut self, parts: TagParts) -> Result<Now, PyParseError> {
+        let mut lexer = NowLexer::new(self.template, parts.clone());
+
+        let format_token = match lexer.next() {
+            Some(res) => res.map_err(ParseError::from)?,
+            None => {
+                return Err(ParseError::from(NowError::Syntax {
+                    at: (parts.at.0 + parts.at.1, 0).into(),
+                })
+                .into());
```
LilyFirefly: I think it would make sense to add a `ParseError::MissingArgument` error (similar to `ParseError::MissingArguments`) for this case, which can be reused for other tags.

dtl-lexer/src/tag/now.rs
```diff
@@ -0,0 +1,107 @@
...
+            seen_as: false,
+            seen_variable: false,
+        }
+    }
+}
+
+impl<'t> Iterator for NowLexer<'t> {
+    type Item = Result<NowToken, NowError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
LilyFirefly: It might make the parsing code tidier if we refactor this into `lex_format` and `lex_variable` methods.

`lex_format` could return `Result<At, NowError>` and `lex_variable` could return `Result<Option<At>, NowError>`.

src/parse.rs
```diff
@@ -1329,6 +1335,52 @@ impl<'t, 'py> Parser<'t, 'py> {
...
     }
 
+    fn parse_now(&mut self, parts: TagParts) -> Result<Now, PyParseError> {
+        let mut lexer = NowLexer::new(self.template, parts.clone());
+
+        let format_token = match lexer.next() {
+            Some(res) => res.map_err(ParseError::from)?,
+            None => {
+                return Err(ParseError::from(NowError::Syntax {
+                    at: (parts.at.0 + parts.at.1, 0).into(),
```
LilyFirefly: Is there a reason this can't be:

```suggestion
                    at: parts.at.into(),
```

src/parse.rs
```diff
@@ -1329,6 +1335,52 @@ impl<'t, 'py> Parser<'t, 'py> {
...
+        }
+
+        if let Some(extra) = lexer.next() {
+            return Err(ParseError::from(extra.unwrap_err()).into());
+        }
+
+        Ok(Now {
+            format_at: format_token.at,
+            asvar,
+        })
```
LilyFirefly, line 1405: At this point, I think `format_at` should become a `TagElement` so it can be resolved as normal in `render`. We should also have tests for the format being a variable instead of a string.

tests/tags/test_now.py
```diff
@@ -0,0 +1,210 @@
...
+from django.utils.dateformat import format as django_format
+from django.test import override_settings
+
+
+def test_now_01(assert_render):
+    expected = "%d %d %d" % (
+        datetime.now().day,
+        datetime.now().month,
+        datetime.now().year,
+    )
```
LilyFirefly: We should be able to use [`strftime`](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior) here.

We'll also probably want to use [`time-machine`](https://time-machine.readthedocs.io/en/latest/) to avoid date/time test flakiness, so long as we're delegating to Django's implementation. As written, this test will be flaky at midnight.

If we ever reimplement these formatting options in Rust, we'll probably be left needing to parse the rendered output and asserting it's close to the expected time, instead of using `time-machine`, since that only patches Python date/time functions, not Rust functions.

tests/tags/test_now.py
```diff
@@ -0,0 +1,210 @@
+from datetime import datetime
+from django.utils import timezone
+from django.utils.formats import date_format
+from django.utils.dateformat import format as django_format
+from django.test import override_settings
+
+
+def test_now_01(assert_render):
```
LilyFirefly: It would be nice to have more meaningful test names if possible.

src/render/tags.rs
```diff
@@ -1227,3 +1232,53 @@ impl Render for SimpleBlockTag {
...
+
+        let tz_mod = DJANGO_TIMEZONE
+            .get_or_try_init(py, || -> Result<Py<PyAny>, PyRenderError> {
+                Ok(py.import("django.utils.timezone")?.into())
+            })?
+            .bind(py);
+        let now_dt = tz_mod.call_method0("now")?;
+
+        let use_named_logic =
+            format.is_empty() || format.chars().all(|c| c.is_ascii_uppercase() || c == '_');
```
LilyFirefly: Is this for detecting [`DATE_FORMAT`](https://docs.djangoproject.com/en/5.2/ref/settings/#std-setting-DATE_FORMAT), [`DATETIME_FORMAT`](https://docs.djangoproject.com/en/5.2/ref/settings/#std-setting-DATETIME_FORMAT), [`SHORT_DATE_FORMAT`](https://docs.djangoproject.com/en/5.2/ref/settings/#std-setting-SHORT_DATE_FORMAT) or [`SHORT_DATETIME_FORMAT`](https://docs.djangoproject.com/en/5.2/ref/settings/#std-setting-SHORT_DATETIME_FORMAT)? If so, I think it would be cleaner to use an explicit list of possibilities here.

src/render/tags.rs
```diff
@@ -1227,3 +1232,53 @@ impl Render for SimpleBlockTag {
...
+                })?
+                .bind(py);
+            df_mod.call_method1("format", (now_dt, format))?
+        };
+
+        let rendered = result.cast::<PyString>().map_err(PyErr::from)?;
+
+        if let Some(asvar_at) = self.asvar {
+            let var_name = template.content(asvar_at);
+            context.insert(var_name.to_string(), rendered.clone().into_any());
```
LilyFirefly: Do we need the `.clone()` here?

```suggestion
            context.insert(var_name.to_string(), rendered.into_any());
```

tests/tags/test_now.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+def test_now_too_many_args(assert_parse_error):
+    template = '{% now "j n Y" extra %}'
+    rusty_message = """\
+  × 'now' statement takes one argument
+   ╭────
+ 1 │ {% now "j n Y" extra %}
+   ·                ──┬──
+   ·                  ╰── here
```
LilyFirefly: I think the error message should say something more like `Unexpected argument after format string`. We could also suggest adding the `as` keyword.

tests/tags/test_now.py
```diff
@@ -0,0 +1,210 @@
...
+    template = '{% now "j n Y" as %}'
+    assert_parse_error(
+        template=template,
+        django_message="'now' statement takes one argument",
+        rusty_message="""\
+  × 'now' statement takes one argument
+   ╭────
+ 1 │ {% now "j n Y" as %}
+   ·                ─┬
+   ·                 ╰── here
```
LilyFirefly: In this case, we could suggest that the expected variable name after `as` is missing.

tests/tags/test_now.py
```diff
@@ -0,0 +1,210 @@
...
+    assert_parse_error(
+        template=template,
+        django_message="'now' statement takes one argument",
+        rusty_message="""\
+  × 'now' statement takes one argument
+   ╭────
+ 1 │ {% now "j n Y" as x y %}
+   ·                     ┬
+   ·                     ╰── here
+   ╰────
```
LilyFirefly: We could say something like `unexpected argument after variable` and suggest removing it.

tests/tags/test_now.py
UnknownPlatypus: I think it would be great to implement in rust at some point since it could also be used in the date filter.

We could probably leverage [jiff](https://github.com/BurntSushi/jiff) that is a great lib for datetime handling

p-r-a-v-i-n: > As written, this test will be flaky at midnight.

I'm learning this for first time. thanks for this.

tests/tags/test_now.py
```diff
@@ -0,0 +1,454 @@
...
+        exception=RuntimeError,
+        django_message="cached failure",
+        rusty_message="cached failure",
+    )
+
+
+@time_machine.travel(FIXED_TIME)
+def test_now_weird_mixed_quotes(assert_render):
+    assert_render(
+        template="{% now \"'Y'\" %}",
```
LilyFirefly: We can avoid needing to escape `"` by using `"""`:

```suggestion
        template="""{% now "'Y'" %}""",
```

tests/tags/test_now.py
```diff
@@ -0,0 +1,454 @@
...
+
+
+@time_machine.travel(FIXED_TIME)
+def test_now_cached_formatter_failure(monkeypatch, assert_render_error):
+    import django.utils.formats
+
+    def bad_date_format(*args, **kwargs):
+        raise RuntimeError("cached failure")
+
+    monkeypatch.setattr(django.utils.formats, "date_format", bad_date_format)
```
LilyFirefly: Ooh! Today I learned about [`monkeypatch`](https://docs.pytest.org/en/stable/how-to/monkeypatch.html)!

tests/tags/test_now.py
```diff
@@ -0,0 +1,454 @@
...
+    assert_render_error(
+        template="{% now 'Y' %}",
+        context={},
+        exception=TypeError,
+        django_message="sequence item 0: expected str instance, int found",
+        rusty_message="'int' object cannot be cast as 'str'",
+    )
+
+
+@time_machine.travel(FIXED_TIME)
```
LilyFirefly, line 461: This test doesn't `assert` anything based on this time, so we don't need this:

```suggestion
```

tests/tags/test_now.py
LilyFirefly: I'm not entirely sure whether its worth testing this - I don't expect Django to actually raise an error here in practice. That said, I don't want to `.expect()` the error on our side, so maybe it's worth it for the coverage.

tests/tags/test_now.py
```diff
@@ -0,0 +1,454 @@
...
+def test_now_unknown_named_format_literal(assert_render):
+    assert_render(
+        template="{% now 'NOT_A_REAL_FORMAT' %}",
+        context={},
+        expected="Jan.+0000UTC_PM_RJanuaryPMFalse_January+0000RJanPMUTC",
+    )
+
+
+@time_machine.travel(FIXED_TIME)
+def test_now_non_string_result(monkeypatch, assert_render_error):
```
LilyFirefly: I'm also unsure if we want this test, for the same reasons as `test_now_cached_formatter_failure`.

tests/tags/test_now.py
```diff
@@ -0,0 +1,454 @@
...
+    )
+
+
+@time_machine.travel(FIXED_TIME)
+def test_now_double_quote_incomplete(assert_render):
+    assert_render(
+        template='{% now " %}',
+        context={},
+        expected=date_format(FIXED_TIME, "DATE_FORMAT"),
+    )
```
LilyFirefly, line 228: Ooh, these are weird! I'd have expected Django to raise a `TemplateSyntaxError` here.

tests/tags/test_now.py
```diff
@@ -0,0 +1,454 @@
...
+        context={},
+        expected=expected.strip(),
+    )
+
+
+@override_settings(USE_TZ=True, TIME_ZONE="UTC")
+def test_now_timezone_aware(assert_render):
+    now = timezone.now()
+    expected = django_format(now, "H")
+    assert_render(template="{% now 'H' %}", context={}, expected=expected)
```
LilyFirefly, line 203: This should use `time_machine`.

tests/tags/test_now.py
```diff
@@ -0,0 +1,454 @@
...
+
+@time_machine.travel(FIXED_TIME)
+def test_now_very_long_format(assert_render):
+    long_format = "Y " * 1000
+    year = FIXED_TIME.strftime("%Y")
+    expected = (year + " ") * 1000
+    assert_render(
+        template=f'{{% now "{long_format.strip()}" %}}',
+        context={},
+        expected=expected.strip(),
```
LilyFirefly: We can avoid needing `.strip()` by using `.join()`:

```suggestion
    long_format = " ".join(["Y"] * 1000)
    year = FIXED_TIME.strftime("%Y")
    expected = " ".join([year] * 1000)
    assert_render(
        template=f'{{% now "{long_format}" %}}',
        context={},
        expected=expected,
```

dtl-lexer/src/tag/now.rs
```diff
@@ -0,0 +1,127 @@
...
+                LexerError::IncompleteString { at } => {
+                    let at_tuple = (at.offset(), at.len());
+                    Ok(Some(TagElementToken {
+                        at: at_tuple,
+                        token_type: Variable,
+                    }))
+                }
+                _ => Err(err.into()),
+            },
+        }
```
LilyFirefly: We can use [`transpose()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.transpose) to turn `Option<Result<>>` into `Result<Option<>>`:

```suggestion
        self.lexer.next().transpose().map_err(|err| {
            match err {
                LexerError::IncompleteString { at } => {
                    let at_tuple = (at.offset(), at.len());
                    Ok(Some(TagElementToken {
                        at: at_tuple,
                        token_type: Variable,
                    }))
                }
                _ => Err(err.into()),
            }
        })
```

dtl-lexer/src/tag/now.rs
```diff
@@ -0,0 +1,127 @@
...
+pub enum NowError {
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    LexerError(#[from] LexerError),
+
+    #[error("Unexpected argument after format string")]
+    #[diagnostic(help("If you want to store the result in a variable, use the 'as' keyword."))]
+    UnexpectedAfterFormat {
+        #[label("unexpected argument")]
+        _at: SourceSpan,
```
LilyFirefly: We can avoid needing these `_` prefixes with `#![expect(unused_assignments)]` at the top of the file.

src/parse.rs
```diff
@@ -1341,6 +1347,16 @@ impl<'t, 'py> Parser<'t, 'py> {
         }
     }
 
+    fn parse_now(&mut self, parts: TagParts) -> Result<Now, PyParseError> {
+        let mut lexer = NowLexer::new(self.template, parts.clone());
```
LilyFirefly: I don't think we need this `.clone()` now:

```suggestion
        let mut lexer = NowLexer::new(self.template, parts);
```

tests/tags/test_now.py
LilyFirefly, line 1: We need a test for `{% now format %}` with `context = {"format": "j n Y"}` (or some other format string).

src/render/tags.rs
```diff
@@ -1227,3 +1232,64 @@ impl Render for SimpleBlockTag {
...
+        py: Python<'_>,
+        template: TemplateString<'t>,
+        context: &mut Context,
+    ) -> RenderResult<'t> {
+        let raw = template.content(self.format_at);
+        let format = if raw.len() >= 2 {
+            &raw[1..raw.len() - 1]
+        } else {
+            ""
+        };
```
LilyFirefly: This feels very low-level for the rendering stage. Maybe we can have a `NowFormat` enum that we create in the parsing stage?

src/render/tags.rs
```diff
@@ -1227,3 +1232,64 @@ impl Render for SimpleBlockTag {
...
+            df_mod.call_method1("format", (now_dt, format))?
+        };
+
+        if let Some(asvar_at) = self.asvar {
+            let var_name = template.content(asvar_at);
+            context.insert(var_name.to_string(), result);
+            Ok(Cow::Borrowed(""))
+        } else {
+            let rendered = result.cast_into::<PyString>().map_err(PyErr::from)?;
+            Ok(Cow::Owned(rendered.to_str()?.to_owned()))
```
LilyFirefly, line 1288: I think we can use [`to_cow()`](https://pyo3.rs/main/doc/pyo3/types/trait.pystringmethods#tymethod.to_cow) and [`into_owned()`](https://doc.rust-lang.org/std/borrow/enum.Cow.html#method.into_owned) here:

```suggestion
            Ok(rendered.to_cow()?.into_owned())
```

dtl-lexer/src/tag/now.rs
```diff
@@ -0,0 +1,127 @@
...
+            Some(Ok(token)) => Ok(Some(token)),
+            Some(Err(err)) => match err {
+                LexerError::IncompleteString { at } => {
+                    let at_tuple = (at.offset(), at.len());
+                    Ok(Some(TagElementToken {
+                        at: at_tuple,
+                        token_type: Variable,
+                    }))
+                }
+                _ => Err(err.into()),
```
LilyFirefly: We should also have a test that hits this branch. Perhaps `LexerError::InvalidRemainder` would work, with `"j n Y"invalid` or similar as a format argument.

tests/tags/test_now.py
p-r-a-v-i-n: Yes this is weird.

tests/tags/test_now.py
p-r-a-v-i-n: I added those bcs I was trying to increase the coverage.  also i agree with we should not expect error from our side. So may be i should remove these tests.

src/render/tags.rs
p-r-a-v-i-n: Any code suggestion would be helpful

dtl-lexer/src/tag/now.rs
p-r-a-v-i-n: Rust raising error here but django doesn't.
for example 
`template = '{% now "Y"invalid %}' ` in this case django returning `2026"501vp.m.Saturday50` (for me doesn't make sense)

and rust raised

```
× Could not parse the remainder
   ╭────
 1 │ {% now "Y"invalid %}
   ·           ───┬───
   ·              ╰── here
   ╰────
```

what would be a right move here

tests/tags/test_now.py
p-r-a-v-i-n: I have added test but output surprised me from both rusty and django
output is `2026Thu, 08 Jan 2026 12:00:00 +000001p.m.31_vp.m.` from both instead of `8 1 2026`

LilyFirefly: It looks like Django always treats the `format` as a string, but doesn't validate it in any way. It also always trims the first and last character:

 https://github.com/django/django/blob/3ea659dc1b8d72d3cab2de17d28cd9d2d06d6eed/django/template/defaulttags.py#L1168

This explains all the weird edge cases we're seeing here. I think we should report these upstream to Django as a bug. Until it's fixed upstream (Django 6.1 at the earliest) we'll need to match these bugs in our implementation.

p-r-a-v-i-n: i have raised discussion on forum : https://forum.djangoproject.com/t/question-about-now-tag/43815
but it seem like they don't treat these as bug

dtl-lexer/src/tag/now.rs
p-r-a-v-i-n: @LilyFirefly  Should i write rust test for this instead in test_now.py  bcs i think we shouldn't match with django here.

src/render/tags.rs
p-r-a-v-i-n: I would need little help here.

tests/tags/test_now.py
```diff
@@ -424,21 +424,19 @@ def test_now_format_from_context(assert_render):
...
-    assert_render(
-        template=template,
-        context={},
-        expected=f"{FIXED_TIME.year}",
-    )
-
-    # assert_parse_error(
+    """
+    Django outputs = '2026"001vp.m.Thursday00'
+    and Rust raised Template syntax error
```
p-r-a-v-i-n: @LilyFirefly  how do you see this ?

tests/tags/test_now.py
LilyFirefly: In the future, the ["Django Internals" section of the forum](https://forum.djangoproject.com/c/internals/5) would be a better place for questions that come up in PRs here. The "Using Django" / "Getting started" section is much more oriented to justifying the status quo than questioning it, whereas "Django Internals" is more open to rethinking things (though still somewhat reluctant to change things).

tests/tags/test_now.py
LilyFirefly: The test has been updated so this suggestion is outdated.

dtl-lexer/src/tag/now.rs
```diff
@@ -0,0 +1,135 @@
...
+        match self.template.content(token.at) {
+            "as" => {
+                let Some(var) = self.next_element()? else {
+                    let position_after_as = token.at.0 + token.at.1;
+                    return Err(NowError::MissingVariableAfterAs {
+                        at: SourceSpan::new(position_after_as.into(), 0usize),
+                    });
+                };
+
+                check_variable_attrs(self.template.content(var.at), var.at.0)?;
```
LilyFirefly: I'm not entirely sure this is appropriate - we should have a test for `{% now "j n Y" as foo.bar %}`, which I suspect should be an error but I think isn't currently for us.

src/parse.rs
```diff
@@ -1341,6 +1353,24 @@ impl<'t, 'py> Parser<'t, 'py> {
...
+
+        let format_at = lexer.lex_format().map_err(ParseError::from)?;
+        let asvar = lexer.lex_variable().map_err(ParseError::from)?;
+        lexer.extra_token().map_err(ParseError::from)?;
+        let raw = self.template.content(format_at);
+        let format = if raw.len() >= 2 {
+            &raw[1..raw.len() - 1]
+        } else {
+            ""
+        };
```
LilyFirefly, line 1400: I think it's worth a comment here explaining that Django always trims the first and last character of the `format` without extra validation.

src/render/tags.rs
```diff
@@ -12,6 +12,8 @@ use pyo3::types::{PyBool, PyDict, PyList, PyNone, PyString, PyTuple};
 
 use crate::render::lorem::{COMMON_WORDS, paragraphs, words};
 use dtl_lexer::tag::lorem::LoremMethod;
+// use dtl_lexer::tag::now::Now;
```
LilyFirefly:
```suggestion
```

src/render/tags.rs
LilyFirefly: Did you forget this one?

src/render/tags.rs
LilyFirefly: I like how you resolved this by moving it into the parser.

dtl-lexer/src/tag/now.rs
```diff
@@ -0,0 +1,135 @@
...
+                let end_of_junk = at.offset() + at.len();
+
+                let total_len = end_of_junk - start_of_tag;
+
+                Ok(Some(TagElementToken {
+                    at: (start_of_tag, total_len),
+                    token_type: Variable,
+                }))
+            }
+            _ => Err(err.into()),
```
LilyFirefly: Looking at the other possible `LexerError` variants here, I think they should be handled like we handle `IncompleteString` or like `InvalidRemainder`. We'll need some extra tests for those variants (`IncompleteTranslatedString`, `InvalidVariableName` and `MissingTranslatedString`).

tests/tags/test_now.py
```diff
@@ -0,0 +1,527 @@
...
+    template = '{% now "Y"invalid %}'
+    assert_render(
+        template=template,
+        context={},
+        expected=r'2026"001vp.m.Thursday00',
+    )
+
+
+def test_now_invalid_variable_name(assert_render):
+    template = '{% now "j n Y" as foo.bar %}'
```
LilyFirefly: This should try to use the `foo.bar` variable too:

```suggestion
    template = '{% now "j n Y" as foo.bar %}{{ foo.bar }}'
```

tests/tags/test_now.py
```diff
@@ -0,0 +1,527 @@
...
+        template='{% now "Y" as current_year %}Year: {{ current_year }}',
+        context={},
+        expected=f"Year: {FIXED_TIME.year}",
+    )
+
+
+@time_machine.travel(FIXED_TIME)
+def test_now_as_leading_underscore(assert_render):
+    assert_render(
+        template='{% now "Y" as _now %}',
```
LilyFirefly:
```suggestion
        template='{% now "Y" as _now %}{{ _now }}',
```

tests/tags/test_now.py
```diff
@@ -0,0 +1,527 @@
...
+        context={},
+        expected="",
+    )
+
+
+@time_machine.travel(FIXED_TIME)
+def test_now_as_single_underscore(assert_render):
+    assert_render(
+        template='{% now "Y" as _ %}',
```
LilyFirefly:
```suggestion
        template='{% now "Y" as _ %}{{ _ }}',
```

tests/tags/test_now.py
```diff
@@ -0,0 +1,527 @@
...
+        context={},
+        expected="",
+    )
+
+
+def test_now_as_invalid_double_dot(assert_render):
+    template = '{% now "j n Y" as foo..bar %}'
```
LilyFirefly, line 303:
```suggestion
    template = '{% now "j n Y" as foo..bar %}{{ foo..bar }}'
```

src/render/tags.rs
LilyFirefly: Ah, I tried this and it still needed the `Cow::Owned`, so wasn't any better after all.

PR #283, opened 2026-01-01
Add a benchmark for the {% include %} tag within a {% for %} loop

LilyFirefly:
Django has some optimisation for `{% include %}` that might make sense to apply in DRT too, but I want to have a benchmark to demonstrate the improvement if it exists.

### DIFF
--- a/tests/benchmarks/test_include.py
+++ b/tests/benchmarks/test_include.py
@@ -112,3 +112,11 @@ def test_include_loop(benchmark, template_engine_nocache):
     }
 
     benchmark(template.render, context)
+
+
+def test_nested_include_loop(benchmark, template_engine_nocache):
+    template = "{% for _ in count %}{% include 'benchmarks/nested.txt' %}{% endfor %}"
+    template = template_engine_nocache.from_string(template)
+    context = {"count": range(100)}
+
+    benchmark(template.render, context)
--- a/tests/templates/benchmarks/inner.txt
+++ b/tests/templates/benchmarks/inner.txt
@@ -1,0 +1,1 @@
+Some text content.
--- a/tests/templates/benchmarks/nested.txt
+++ b/tests/templates/benchmarks/nested.txt
@@ -1,0 +1,1 @@
+{% for _ in count %}{% include "benchmarks/inner.txt" %}{% endfor %}


LilyFirefly:
Ah! Looking at the Django commit (https://github.com/django/django/commit/a391b17ad24bc5f255a3928c23c158c79004c656) I think the problem might be that the cached template loader is active.

pyproject.toml
```diff
@@ -28,6 +28,7 @@ dev = [
     "pytest",
     "pytest-django",
     "pytest-cov",
+    "pytest-codspeed",
```
UnknownPlatypus, line 31: Could you run uv lock to update the dependencies lock file ? 
It should fix the pre-commit.ci failure

tests/conftest.py
```diff
@@ -14,6 +15,16 @@ def template_engine(request):
     return engines[request.param]
 
 
+@all_engines_nocache
+def template_engine_nocache(request):
+    """
+    Parametrize tests to run against both rusty and django template engines.
```
UnknownPlatypus: We should maybe update here to say this fixture will force usage of the filesystem backend with no cached loader ?

tests/benchmarks/test_include.py
```diff
@@ -0,0 +1,122 @@
...
+        "users": [
+            {"name": "Alice", "age": 35},
+            {"name": "Bob", "age": 21},
+            {"name": "Carla", "age": 15},
+            {"name": "David"},
+            {"name": "Emma", "age": None},
+        ],
+    }
+
+    benchmark(lambda: template.render(context))
```
UnknownPlatypus:
```suggestion
    benchmark(template.render, context)
```
Maybe good to avoid the lambda ? Not sure if it can interfere in some way but maybe?
https://codspeed.io/docs/reference/pytest-codspeed#direct-invocation

pyproject.toml
LilyFirefly: Ah... I don't have `uv` installed.

LilyFirefly: Ah, I can `pip install uv` - good enough.

.github/workflows/codspeed.yml
```diff
@@ -32,8 +33,22 @@ jobs:
...
+        run: uv sync --dev
+
       - name: Run the benchmarks
         uses: CodSpeedHQ/action@972e3437949c89e1357ebd1a2dbc852fcbc57245 # v4.5.1
         with:
           mode: simulation
-          run: cargo codspeed run -p dtl_benchmark
+          run: |
+            cargo codspeed run -p dtl_benchmark
+            uv run pytest tests/ --codspeed
```
LilyFirefly, line 54: It turns out we need to run both types of benchmark within the same `run:` block here or codspeed gets confused and thinks some of the benchmarks are missing.

pyproject.toml
UnknownPlatypus: they have an [installer](https://docs.astral.sh/uv/getting-started/installation/) otherwise

.github/workflows/codspeed.yml
UnknownPlatypus: Good to know

PR #285, opened 2026-01-02
Cache included templates in Context to avoid get/select template

LilyFirefly:
Calling into `get_template` or `select_template` can be expensive when the cached template loader isn't used.

### DIFF
--- a/src/render/tags.rs
+++ b/src/render/tags.rs
@@ -1,5 +1,4 @@
 use std::borrow::Cow;
-use std::collections::hash_map::Entry;
 use std::collections::{HashMap, VecDeque};
 use std::sync::Arc;
 
@@ -22,9 +21,7 @@ use crate::parse::{
     For, IfCondition, Include, IncludeTemplateName, SimpleBlockTag, SimpleTag, Tag, TagElement, Url,
 };
 use crate::path::construct_relative_path;
-use crate::template::django_rusty_templates::{
-    Engine, NoReverseMatch, Template, TemplateDoesNotExist, get_template, select_template,
-};
+use crate::template::django_rusty_templates::{NoReverseMatch, Template, TemplateDoesNotExist};
 use crate::utils::PyResultMethods;
 
 static PROMISE: PyOnceLock<Py<PyAny>> = PyOnceLock::new();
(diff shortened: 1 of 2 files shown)


src/render/tags.rs
```diff
@@ -802,6 +804,26 @@ impl<'t, 'py> IncludeTemplate<'t> {
     }
 }
 
+fn get_include(
+    py: Python,
+    engine: &Arc<Engine>,
+    context: &mut Context,
+    key: IncludeTemplateKey,
+) -> Result<(IncludeTemplateKey, Template), PyErr> {
+    let include = match context.include_cache.remove(&key) {
```
LilyFirefly: I needed to remove the included template from the cache and put it back later to work around the borrow checker. There might be a way to avoid this, but this works and is performant.

src/render/types.rs
```diff
@@ -53,13 +54,20 @@ impl ForLoop {
...
+}
+
 #[derive(Debug, Default)]
 pub struct Context {
     context: HashMap<String, Vec<Py<PyAny>>>,
     loops: Vec<ForLoop>,
     pub request: Option<Py<PyAny>>,
     pub autoescape: bool,
     names: Vec<HashSet<String>>,
+    pub include_cache: HashMap<IncludeTemplateKey, Template>,
```
UnknownPlatypus: I know ruff and others are often using `rustc_hash::FxHashMap` that can squeeze some additional performance. Might be worth to try (maybe in a followup pr)

LilyFirefly: Good idea! I've also got ideas of using [`SmallVec`](https://crates.io/crates/smallvec) in various places (mainly `Context`). But I will leave these ideas to follow-up PRs.

src/render/tags.rs
LilyFirefly: I realised I could use `Arc` to work around the same problem and I'm happier with the resulting code.

src/render/types.rs
UnknownPlatypus: > But I will leave these ideas to follow-up PRs.

Probably better to do when we've setup more benchmarks to see the actual impact of changes.

PR #297, opened 2026-01-08
Fix LoremLexer handling of variables with filters

LilyFirefly:
I realised since merging #276 that we weren't handling filters with arguments containing whitespace correctly in the `{% lorem %}` tag. I'm starting to think most tag lexers should be wrappers around `SimpleTagLexer` to ensure this is always handled correctly.


p-r-a-v-i-n:
We should merge this before #298

PR #317, opened 2026-01-27
Add comment filter

p-r-a-v-i-n:
part #2


src/parse.rs
```diff
@@ -1341,6 +1353,28 @@ impl<'t, 'py> Parser<'t, 'py> {
         }
     }
 
+    fn parse_comment(&mut self, at: At, parts: TagParts) -> Result<Comment, PyParseError> {
+        let lexer = CommentLexer::new(self.template, parts);
```
LilyFirefly: We can just use `TagElementLexer` directly here:

```suggestion
        let lexer = TagElementLexer::new(self.template, parts);
```

tests/tags/test_comment.py
```diff
@@ -0,0 +1,120 @@
...
+"""),
+    )
+
+
+def test_comment_parsing_errors(assert_parse_error):
+    """
+    Test errors in the opening comment tag arguments.
+    """
+    template = '{% comment "unclosed string %}'
+    django_msg = "Unclosed tag on line 1: 'comment'. Looking for one of: endcomment."
```
LilyFirefly: Does the Django error disappear if we add an `{% endcomment %}` tag?

```suggestion
    template = '{% comment "unclosed string %}{% endcomment %}'
    django_msg = "Unclosed tag on line 1: 'comment'. Looking for one of: endcomment."
```

tests/tags/test_comment.py
```diff
@@ -0,0 +1,120 @@
...
+        django_message=django_msg,
+        rusty_message=snapshot("""\
+  × Could not parse the remainder
+   ╭────
+ 1 │ {% comment "note"invalid %}
+   ·                  ───┬───
+   ·                     ╰── here
+   ╰────
+"""),
+    )
```
LilyFirefly, line 115: This should be its own test, not part of `test_comment_parsing_errors`.

tests/tags/test_comment.py
```diff
@@ -0,0 +1,120 @@
...
+ 1 │ {% comment "unclosed string %}
+   ·            ────────┬───────
+   ·                    ╰── here
+   ╰────
+"""),
+    )
+
+    # Invalid remainder (missing space)
+    template = '{% comment "note"invalid %}'
+    django_msg = "Unclosed tag on line 1: 'comment'. Looking for one of: endcomment."
```
LilyFirefly, line 103: I also wonder if adding `{% endcomment %}` here will change the Django behaviour.

src/parse.rs
```diff
@@ -1341,6 +1353,28 @@ impl<'t, 'py> Parser<'t, 'py> {
...
+        for token in lexer {
+            token.map_err(ParseError::from)?;
+        }
```
LilyFirefly: We might not need this at all (see my comments in the Python tests). In which case, we don't need the `CommentLexer` or `TagElementLexer` at all here.

src/parse.rs
```diff
@@ -1341,6 +1353,28 @@ impl<'t, 'py> Parser<'t, 'py> {
...
+                    return Ok(Comment);
+                }
+            }
+        }
+
+        Err(ParseError::UnclosedTag {
+            expected: "endcomment".to_string(),
+            at: at.into(),
+        }
+        .into())
```
LilyFirefly: We can use `MissingEndTag` here instead of introducing this new `UnclosedTag` error variant.

```suggestion
        Err(ParseError::MissingEndTag {
            start: Cow::Borrowed("comment"),
            expected: "endcomment".to_string(),
            at: at.into(),
        }
        .into())
```

It would also be nice to tidy `MissingEndTag` to use `Cow<'static, str>` instead of needing `String` for the `expected` attribute.

src/render/tags.rs
```diff
@@ -709,6 +710,7 @@ impl Render for Tag {
 
                 Cow::Owned(text)
             }
+            Self::Comment(_comment) => render_comment(),
```
LilyFirefly: We can just inline `Cow::Borrowed("")` here and delete `src/render/comment.rs`.

LilyFirefly (commented):
One last suggestion, then this can be merged.

Issue #326, opened 2026-02-02
Consider moving from quickcheck to proptest


LilyFirefly:
While `quickcheck` works and is easy to use, it hasn't had a release in 5 years and is blocking an [upgrade to `rand` `0.9`](https://github.com/LilyFirefly/django-rusty-templates/pull/289). `proptest` continues to have frequent releases, so should be a workable alternative.

PR #329, opened 2026-02-06
Removed quickcheck and add proptest

p-r-a-v-i-n:
Refs #326


proptest-regressions/loaders.txt
```diff
@@ -0,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
```
p-r-a-v-i-n, line 3: this gets autogenerated

PR #341, opened 2026-02-12
Include dtl-lexer in python coverage report

### LINKED ISSUE #340: Work out why dtl-lexer code coverage isn't reported from Python tests
We use `cargo-llvm-cov` to generate Rust coverage, but we need to jump through a bunch of hoops to enable this for code run via Python tests: https://github.com/LilyFirefly/django-rusty-templates/blob/07f8a3e680cf08b9782e60c6b3fa4530efa2a0e3/.github/workflows/python.yml#L73-L93

Unfortunately, the resulting `coverage.json` file doesn't report coverage for the `dtl-lexer` crate, so we only get coverage information from our Rust tests. I haven't been able to track down why this doesn't work correctly.

p-r-a-v-i-n:
When running `cargo llvm-cov `report from the root of a workspace, it defaults to collecting coverage only for the root package `(django-rusty-templates).` Since `dtl-lexer` is a separate crate within the workspace, its coverage data was generated during tests but simply ignored when generating the final report.

- Explicitly include `dtl-lexer` and `django-rusty-templates` packages in `cargo llvm-cov report` command

closes #340

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -90,7 +90,7 @@ jobs:
       run: |
         cargo llvm-cov show-env --sh > llvm-cov-env.sh
         source llvm-cov-env.sh
-        cargo llvm-cov report --codecov --output-path codecov.json
+        cargo llvm-cov report -p dtl-lexer -p django-rusty-templates --codecov --output-path codecov.json
 
     - name: Upload to codecov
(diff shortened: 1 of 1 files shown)


UnknownPlatypus:
Is there a "all workspaces" option ? We will have the same issue if we split something again otherwise

PR #397, opened 2026-04-05
Ignored whitespace before commas in variable names in forloop

### LINKED ISSUE #396: `ForLexer::lex_in` raises `ForLexerInError::MissingComma` on `{% for key , value in items %}`
The following are correctly parsed:

- `{% for key, value in items %}`
- `{% for key,value in items %}`

but the following raises `ForLexerInError::MissingComma`: `{% for key , value in items %}` even though [this is a legit and tested syntax](https://github.com/django/django/blob/main/tests/template_tests/syntax_tests/test_for.py#L103-L114).

p-r-a-v-i-n:
Fixes #396 
- We were seperating the variable names using whitespace inside the `lex_variable_name`. So when we used something like ` key , value in items` , we only get `key` as variable names and when it reaches to `lex_in()` it have `, value in items` in it,  so it raises error their.

### DIFF
--- a/dtl-lexer/src/tag/forloop.rs
+++ b/dtl-lexer/src/tag/forloop.rs
@@ -244,6 +244,11 @@ impl<'t> ForLexer<'t> {
             State::Done => return None,
         }
 
+        let first_non_ws_index = self.rest.next_non_whitespace();
+        if self.rest[first_non_ws_index..].starts_with(',') {
+            let at = (self.byte + first_non_ws_index, 1);
+            return Some(Err(ForLexerError::UnexpectedComma { at: at.into() }));
+        }
         let index = self.rest.next_whitespace();
         let (index, next_index) = match self.rest.find(',') {
             Some(comma_index) if comma_index < index => {
--- a/dtl-lexer/src/tag/forloop.rs
+++ b/dtl-lexer/src/tag/forloop.rs
@@ -236,21 +236,18 @@ impl<'t> ForLexer<'t> {
 
     fn get_index_and_next_index(&mut self, index: usize) -> Result<(usize, usize), ForLexerError> {
         let (index, next_index) = match self.rest.find(',') {
+            Some(0) => {
+                let at = (self.byte, 1);
+                return Err(ForLexerError::UnexpectedComma { at: at.into() });
+            }
             Some(comma_index) if comma_index < index => {
                 let next_index = self.rest[comma_index + 1..].next_non_whitespace();
-                let after_comma = comma_index + 1 + next_index;
-                if after_comma < self.rest.len() && self.rest[after_comma..].starts_with(',') {
-                    let at = (self.byte + after_comma, 1);
-                    return Err(ForLexerError::UnexpectedExpression { at: at.into() });
-                }
                 (comma_index, comma_index + 1 + next_index)
             }
             _ => {
                 let whitespace_to_next = self.rest[index..].next_non_whitespace();
                 let after_whitespace = index + whitespace_to_next;
-                if after_whitespace < self.rest.len()
-                    && self.rest[after_whitespace..].starts_with(',')
-                {
+                if self.rest[after_whitespace..].starts_with(',') {
                     let next_index = self.rest[after_whitespace + 1..].next_non_whitespace();
                     let after_comma = after_whitespace + 1 + next_index;
                     if after_comma < self.rest.len() && self.rest[after_comma..].starts_with(',') {
@@ -299,11 +296,6 @@ impl<'t> ForLexer<'t> {
             State::Done => return None,
         }
 
-        let first_non_ws_index = self.rest.next_non_whitespace();
-        if self.rest[first_non_ws_index..].starts_with(',') {
-            let at = (self.byte + first_non_ws_index, 1);
-            return Some(Err(ForLexerError::UnexpectedComma { at: at.into() }));
-        }
         let index = self.rest.next_whitespace();
         let (index, next_index) = match self.get_index_and_next_index(index) {
             Ok(indexes) => indexes,
--- a/dtl-lexer/src/tag/forloop.rs
+++ b/dtl-lexer/src/tag/forloop.rs
@@ -268,6 +268,27 @@ impl<'t> ForLexer<'t> {
         Ok((index, next_index))
     }
 
+    fn check_invalid_variable_name(&mut self, index: usize, at: At) -> Result<(), ForLexerError> {
+        let name = &self.rest[..index];
+
+        let (_, _, remainder) = lex_variable(0, name);
+        if !remainder.is_empty() {
+            return Err(ForLexerError::InvalidName {
+                name: name.to_string(),
+                at: at.into(),
+            });
+        }
+
+        if name == '.'.to_string() || name.contains(['"', '\'', '|']) {
+            return Err(ForLexerError::InvalidName {
+                name: name.to_string(),
+                at: at.into(),
+            });
+        }
+
+        Ok(())
+    }
+
     pub fn lex_variable_name(&mut self) -> Option<Result<ForVariableNameToken, ForLexerError>> {
         match self.state {
             State::VariableName if !self.rest.is_empty() => {}
@@ -290,25 +311,13 @@ impl<'t> ForLexer<'t> {
         };
         let at = (self.byte, index);
         self.previous_at = Some(at);
-        let name = &self.rest[..index];
 
-        if name == ';'.to_string() || name == '.'.to_string() {
+        if let Err(err) = self.check_invalid_variable_name(index, at) {
             self.rest = "";
             self.state = State::Done;
-            return Some(Err(ForLexerError::InvalidName {
-                name: name.to_string(),
-                at: at.into(),
-            }));
+            return Some(Err(err));
         }
 
-        if name.contains(['"', '\'', '|']) {
-            self.rest = "";
-            self.state = State::Done;
-            return Some(Err(ForLexerError::InvalidName {
-                name: name.to_string(),
-                at: at.into(),
-            }));
-        }
         self.byte += next_index;
         self.rest = &self.rest[next_index..];
         Some(Ok(ForVariableNameToken { at }))
--- a/dtl-lexer/src/tag/forloop.rs
+++ b/dtl-lexer/src/tag/forloop.rs
@@ -234,31 +234,6 @@ impl<'t> ForLexer<'t> {
         Err(ForLexerError::UnexpectedExpression { at: at.into() })
     }
 
-    fn get_index_and_next_index(&mut self, index: usize) -> Result<(usize, usize), ForLexerError> {
-        let (index, next_index) = match self.rest.find(',') {
-            Some(0) => {
-                let at = (self.byte, 1);
-                return Err(ForLexerError::UnexpectedComma { at: at.into() });
-            }
-            Some(comma_index) if comma_index < index => {
-                let next_index = self.rest[comma_index + 1..].next_non_whitespace();
-                (comma_index, comma_index + 1 + next_index)
-            }
-            _ => {
-                let after_whitespace = index + self.rest[index..].next_non_whitespace();
-                if self.rest[after_whitespace..].starts_with(',') {
-                    let next_index = self.rest[after_whitespace + 1..].next_non_whitespace();
-                    (index, after_whitespace + 1 + next_index)
-                } else {
-                    self.state = State::Done;
-                    (index, after_whitespace)
-                }
-            }
-        };
-
-        Ok((index, next_index))
-    }
-
     fn check_invalid_variable_name(&mut self, index: usize, at: At) -> Result<(), ForLexerError> {
         let name = &self.rest[..index];
 
@@ -291,9 +266,25 @@ impl<'t> ForLexer<'t> {
         }
 
         let index = self.rest.next_whitespace();
-        let (index, next_index) = match self.get_index_and_next_index(index) {
-            Ok(indexes) => indexes,
-            Err(e) => return Some(Err(e)),
+        let (index, next_index) = match self.rest.find(',') {
+            Some(0) => {
+                let at = (self.byte, 1);
+                return Some(Err(ForLexerError::UnexpectedComma { at: at.into() }));
+            }
+            Some(comma_index) if comma_index < index => {
+                let next_index = self.rest[comma_index + 1..].next_non_whitespace();
+                (comma_index, comma_index + 1 + next_index)
+            }
+            _ => {
+                let after_whitespace = index + self.rest[index..].next_non_whitespace();
+                if self.rest[after_whitespace..].starts_with(',') {
+                    let next_index = self.rest[after_whitespace + 1..].next_non_whitespace();
+                    (index, after_whitespace + 1 + next_index)
+                } else {
+                    self.state = State::Done;
+                    (index, after_whitespace)
+                }
+            }
         };
         let at = (self.byte, index);
         self.previous_at = Some(at);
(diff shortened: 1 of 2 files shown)


tests/tags/test_for.py
```diff
@@ -795,3 +795,17 @@ def test_missing_argument_after_for_loop(assert_render_error):
...
+    context = {"items": (("one", 1), ("two", 2))}
+    expected = "one:1/two:2/"
+    assert_render(template=template, context=context, expected=expected)
+
+
+def test_for_tag_unpack05(assert_render):
+    template = "{% for num1, num2, num3, num4 in items %}{{ num1 }}{{ num2 }}{{ num3 }}{{ num4 }}/{% endfor %}"
+    context = {"items": ((1, 2, 3, 4), (4, 5, 6, 7))}
+    expected = "1234/4567/"
+    assert_render(template=template, context=context, expected=expected)
```
p-r-a-v-i-n: In this case rusty did not raised error but the output it generate as ` 13/46/`

tests/tags/test_for.py
```diff
@@ -823,3 +823,66 @@ def test_for_tag_with_more_variable_names_without_whitespace(assert_render):
...
+        "'for' tag received an invalid argument: for key,,value in items"
+    )
+    rusty_message = snapshot("""\
+  × Unexpected expression in for loop:
+   ╭────
+ 1 │ {% for key,,value in items %}{{ key }}:{{ value }}/{% endfor %}
+   ·            ┬
+   ·            ╰── unexpected expression
+   ╰────
+""")
```
LilyFirefly: Perhaps we can make this:

```suggestion
  × Unexpected comma in for loop:
   ╭────
 1 │ {% for key,,value in items %}{{ key }}:{{ value }}/{% endfor %}
   ·            ┬
   ·            ╰── here
   ╰────
  help: Try removing the comma, or adding a variable name before it
""")
```

tests/tags/test_for.py
```diff
@@ -795,3 +795,94 @@ def test_missing_argument_after_for_loop(assert_render_error):
...
+    assert_parse_error(
+        template=template, django_message=django_message, rusty_message=rusty_message
+    )
+
+
+def test_for_tag_unpack08(assert_parse_error):
+    """
+    This test were taken from django.
+    """
+    template = "{% for key,value, in items %}{{ key }}:{{ value }}/{% endfor %}"
```
LilyFirefly, line 873: We should also add a test like:

```python
    template = "{% for ,value in items %}{{ value }}/{% endfor %}"
```

dtl-lexer/src/tag/forloop.rs
```diff
@@ -237,16 +237,35 @@ impl<'t> ForLexer<'t> {
...
+                let whitespace_to_next = self.rest[index..].next_non_whitespace();
+                let after_whitespace = index + whitespace_to_next;
+                if after_whitespace < self.rest.len()
+                    && self.rest[after_whitespace..].starts_with(',')
+                {
+                    let next_index = self.rest[after_whitespace + 1..].next_non_whitespace();
+                    let after_comma = after_whitespace + 1 + next_index;
+                    if after_comma < self.rest.len() && self.rest[after_comma..].starts_with(',') {
+                        let at = (self.byte + after_comma, 1);
+                        return Some(Err(ForLexerError::UnexpectedExpression { at: at.into() }));
```
LilyFirefly: These are the currently uncovered lines. (I also got the bad request until I logged into codecov, so maybe that'll help you too.)

tests/tags/test_for.py
LilyFirefly: Some more tests that I'd be interested in adding would be something like:

```python
    template = "{% for key ; value in items %}{{ key }}:{{ value }}/{% endfor %}"
```
```python
    template = "{% for key , ; value in items %}{{ key }}:{{ value }}/{% endfor %}"
```

dtl-lexer/src/tag/forloop.rs
p-r-a-v-i-n: thanks

tests/tags/test_for.py
p-r-a-v-i-n: these were critical indeed. I have added them. Thanks :)

tests/tags/test_for.py
```diff
@@ -886,3 +905,58 @@ def test_for_tag_without_comma(assert_parse_error):
...
+    template = "{% for key , ; value in items %}{{ key }}:{{ value }}/{% endfor %}"
+    django_message = snapshot(
+        "'for' tag received an invalid argument: for key , ; value in items"
+    )
+    rusty_message = snapshot("""\
+  × Unexpected expression in for loop. Did you miss a comma when unpacking?
+   ╭────
+ 1 │ {% for key , ; value in items %}{{ key }}:{{ value }}/{% endfor %}
+   ·                ──┬──
+   ·                  ╰── unexpected expression
```
LilyFirefly: I think we should try to do better here - I don't believe a `;` is a valid variable name, so we should flag that instead.

p-r-a-v-i-n: you are right. 
should we create a another Error formatter ? so it will flag something like `unexpected token ; after comma`

LilyFirefly: Yeah, sounds right. I'd probably say something like `Invalid variable name` though - this one should trigger for anything where we have invalid characters. `;` was just the example I thought of, it's not special (other than looking a bit like a comma).

dtl-lexer/src/tag/forloop.rs
```diff
@@ -279,6 +264,30 @@ impl<'t> ForLexer<'t> {
                 }
             }
         };
+
+        Ok((index, next_index))
+    }
```
p-r-a-v-i-n: I don't mind if this get reverted. IMO this will add more readability.

dtl-lexer/src/tag/forloop.rs
```diff
@@ -237,21 +277,30 @@ impl<'t> ForLexer<'t> {
...
-            }
+        let (index, next_index) = match self.get_index_and_next_index(index) {
+            Ok(indexes) => indexes,
+            Err(e) => return Some(Err(e)),
         };
         let at = (self.byte, index);
         self.previous_at = Some(at);
         let name = &self.rest[..index];
+
+        if name == ';'.to_string() || name == '.'.to_string() {
```
LilyFirefly: There's a lot more invalid names possible. We're going to need something a lot more generic. See if you can use [`lex_variable`](https://github.com/LilyFirefly/django-rusty-templates/blob/279d80c1af6e277b4f78c8bd548fc37dd6c076c1/dtl-lexer/src/common.rs#L53) here? That might not be a perfect fit because it's designed for `dotted.variables` too.

tests/tags/test_for.py
```diff
@@ -996,3 +996,57 @@ def test_for_tag_invalid_variable_names_02(assert_parse_error):
...
+  × Unexpected expression in for loop. Did you miss a comma when unpacking?
+   ╭────
+ 1 │ {% for key , yiuyi value in items %}{{ key }}:{{ value }}/{% endfor %}
+   ·                    ──┬──
+   ·                      ╰── unexpected expression
+   ╰────
+""")
+    assert_parse_error(
+        template=template, django_message=django_message, rusty_message=rusty_message
+    )
```
p-r-a-v-i-n, line 1035: Are we correctly handling this case ?

LilyFirefly (commented):
Can we try some tests with variable names like `ab;%t$s`, `foo.bar`, `foo|upper`, and `foo|default:"bar"`? I think all of these should be invalid (but check Django agrees).

tests/tags/test_for.py
LilyFirefly: Looks correct to me - adding a comma between `yiuyi` and `value` would make it syntactically correct.

p-r-a-v-i-n:
> Can we try some tests with variable names like `ab;%t$s`, `foo.bar`, `foo|upper`, and `foo|default:"bar"`? I think all of these should be invalid (but check Django agrees).

sure

p-r-a-v-i-n:
Django allowing random token in the variable name for example
`foo.bar` , `foo;bar`,, `foo$$bar` , `foo$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$bar`, `foo$@$;%t$$bar` is also valid variable name  in django

LilyFirefly:
> Django allowing random token in the variable name for example `foo.bar` , `foo;bar`,, `foo$$bar` , `foo$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$bar`, `foo$@$;%t$$bar` is also valid variable name in django

That surprises me, but good to know! Are these usable as variables too (`{{ foo.bar }}`, `{{ foo;bar }}` etc)? If so we should just support this directly. If not, I think it's time for another bug report to Django.

LilyFirefly:
Ah, it's an awkward middle ground where it renders something, but almost certainly not what the developer intended. Probably still worth a bug report.

p-r-a-v-i-n:
> Ah, it's an awkward middle ground where it renders something, but almost certainly not what the developer intended. Probably still worth a bug report.

please wait for few minutes, i think i have messup something in the tests

p-r-a-v-i-n:
> > Django allowing random token in the variable name for example `foo.bar` , `foo;bar`,, `foo$$bar` , `foo$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$bar`, `foo$@$;%t$$bar` is also valid variable name in django
> 
> That surprises me, but good to know! Are these usable as variables too (`{{ foo.bar }}`, `{{ foo;bar }}` etc)? If so we should just support this directly. If not, I think it's time for another bug report to Django.

Sorry ,  I think i might have  missed to run `uv run maturin develop` before running the test .  now they are behaving as expected.

tests/tags/test_for.py
```diff
@@ -795,3 +795,412 @@ def test_missing_argument_after_for_loop(assert_render_error):
...
+
+
+def test_for_tag_variable_name_with_dot_in_unpack(assert_render):
+    template = "{% for foo.bar in items %}{{ foo.bar }}{% endfor %}"
+    context = {"items": ["django", "rusty", "templates"]}
+    expected = ""
```
p-r-a-v-i-n, line 1075: This could be questionable ?

LilyFirefly: Since that's what Django does, we should too. However, I think opening a bug report in Django makes sense, since this behaviour is almost certainly not what the developer intended.

p-r-a-v-i-n: You are right.

dtl-lexer/src/tag/forloop.rs
```diff
@@ -228,6 +234,61 @@ impl<'t> ForLexer<'t> {
...
+                    let at = (self.byte + after_comma, 1);
+                    return Err(ForLexerError::UnexpectedExpression { at: at.into() });
+                }
+                (comma_index, comma_index + 1 + next_index)
+            }
+            _ => {
+                let whitespace_to_next = self.rest[index..].next_non_whitespace();
+                let after_whitespace = index + whitespace_to_next;
+                if after_whitespace < self.rest.len()
+                    && self.rest[after_whitespace..].starts_with(',')
```
LilyFirefly: We don't need to check against the `len` here - given how we're constructing these indexes, that's guaranteed to be valid.

```suggestion
                if self.rest[after_whitespace..].starts_with(',')
```

dtl-lexer/src/tag/forloop.rs
```diff
@@ -237,31 +298,28 @@ impl<'t> ForLexer<'t> {
             }
             State::Done => return None,
         }
+
+        let first_non_ws_index = self.rest.next_non_whitespace();
+        if self.rest[first_non_ws_index..].starts_with(',') {
+            let at = (self.byte + first_non_ws_index, 1);
+            return Some(Err(ForLexerError::UnexpectedComma { at: at.into() }));
+        }
```
LilyFirefly: I think we can inline this into `get_index_and_next_index`.


```suggestion
```

dtl-lexer/src/tag/forloop.rs
```diff
@@ -228,6 +234,61 @@ impl<'t> ForLexer<'t> {
         Err(ForLexerError::UnexpectedExpression { at: at.into() })
     }
 
+    fn get_index_and_next_index(&mut self, index: usize) -> Result<(usize, usize), ForLexerError> {
+        let (index, next_index) = match self.rest.find(',') {
+            Some(comma_index) if comma_index < index => {
```
LilyFirefly: We can inline the check for a leading `comma` here I think:

```suggestion
        let (index, next_index) = match self.rest.find(',') {
            Some(0) => {
                let at = (self.byte, 1);
                return Some(Err(ForLexerError::UnexpectedComma { at: at.into() }));
            }
            Some(comma_index) if comma_index < index => {
```

dtl-lexer/src/tag/forloop.rs
```diff
@@ -228,6 +234,61 @@ impl<'t> ForLexer<'t> {
...
+                let next_index = self.rest[comma_index + 1..].next_non_whitespace();
+                let after_comma = comma_index + 1 + next_index;
+                if after_comma < self.rest.len() && self.rest[after_comma..].starts_with(',') {
+                    let at = (self.byte + after_comma, 1);
+                    return Err(ForLexerError::UnexpectedExpression { at: at.into() });
+                }
```
LilyFirefly: With the leading comma check inlined above, I think we no longer need this:

```suggestion
```

dtl-lexer/src/tag/forloop.rs
```diff
@@ -228,6 +234,61 @@ impl<'t> ForLexer<'t> {
...
+                let after_whitespace = index + whitespace_to_next;
+                if after_whitespace < self.rest.len()
+                    && self.rest[after_whitespace..].starts_with(',')
+                {
+                    let next_index = self.rest[after_whitespace + 1..].next_non_whitespace();
+                    let after_comma = after_whitespace + 1 + next_index;
+                    if after_comma < self.rest.len() && self.rest[after_comma..].starts_with(',') {
+                        let at = (self.byte + after_comma, 1);
+                        return Err(ForLexerError::UnexpectedComma { at: at.into() });
+                    }
```
LilyFirefly: If we inline the leading comma check above, I think we also don't need this:

```suggestion
```

LilyFirefly (commented):
Great work, this is getting really close!

I think we can simplify the logic around leading commas. When I tried making these changes locally, one of the tests had the error message change, but I think it became better.

dtl-lexer/src/tag/forloop.rs
LilyFirefly: I checked and we still don't need this.

dtl-lexer/src/tag/forloop.rs
```diff
@@ -228,6 +234,59 @@ impl<'t> ForLexer<'t> {
...
+    fn get_index_and_next_index(&mut self, index: usize) -> Result<(usize, usize), ForLexerError> {
+        let (index, next_index) = match self.rest.find(',') {
+            Some(0) => {
+                let at = (self.byte, 1);
+                return Err(ForLexerError::UnexpectedComma { at: at.into() });
+            }
+            Some(comma_index) if comma_index < index => {
+                let next_index = self.rest[comma_index + 1..].next_non_whitespace();
+                let after_comma = comma_index + 1 + next_index;
+                (comma_index, after_comma)
```
LilyFirefly: We can merge these lines:

```suggestion
                (comma_index, comma_index + 1 + next_index)
```

dtl-lexer/src/tag/forloop.rs
LilyFirefly: I think once you've applied the rest of my suggestions, this will be small enough to inline again.

Issue #1, opened 2024-10-29
Implement all builtin filters


LilyFoote:
We need to support every filter in https://docs.djangoproject.com/en/5.1/ref/templates/builtins/#ref-templates-builtins-filters. They can be added as variants to the `Filter` enum in `src/parse.rs`.

jonathan-s:
Making an attempt at addslashes.

jonathan-s:
I've been thinking about filters and how they are structured. In `parse.rs` we get the filter type. Then we implement render for filter, and thus match it on filter type, and returning the correct behaviour there. We get a very long match statement there, and it has a tendency to get a bit unwieldy. 

An idea that could perhaps work would be to use unitstructs in `FilterType`. 

```rust
pub struct AddSlashesFilter;

pub enum Filtertype {
  AddSlashes(AddSlashesFilter)
  ..
}

```

Given we can then implement a trait `apply` for `filter`, and each filter would implement that. I think that could possibly be a pretty nice way of decoupling and breaking up that big match statement.

LilyFoote:
I think that's a great idea, go for it!

I've also been wondering defining a macro (or two) to abstract away the repetitive error handling in the big `match` in `parse.rs`.

PR #409, opened 2026-04-18
Add `force_escape` filter

Mouarius:
Adds the `force_escape` template filter (see https://github.com/LilyFirefly/django-rusty-templates/issues/1) support by forcing the call to `EscapeFilter.resolve()` even if the content is already `ContentString::HtmlSave`.

The python tests are willingly repetitive to follow Django tests definition. I found using a small wrapper function for `autoescape off` block not really useful in this case.

Let me know what you think !

### DIFF
--- a/src/render/filters.rs
+++ b/src/render/filters.rs
@@ -497,6 +497,13 @@ impl ResolveFilter for ForceEscapeFilter {
         template: TemplateString<'t>,
         context: &mut Context,
     ) -> ResolveResult<'t, 'py> {
+        let variable = match variable {
+            // We want to re-run the escape filter even if the content has already been resolved to safe HTML
+            Some(Content::String(ContentString::HtmlSafe(c))) => {
+                Some(Content::String(ContentString::String(c)))
+            }
+            other => other,
+        };
         EscapeFilter.resolve(variable, py, template, context)
     }
 }
--- a/src/render/filters.rs
+++ b/src/render/filters.rs
@@ -217,10 +217,10 @@ impl ResolveFilter for CutFilter {
         let result = match content_string {
             ContentString::String(s) => ContentString::String(cut(s, &arg).into()),
             ContentString::HtmlSafe(s) => {
-                let cut = cut(s, &arg);
+                let cut = cut(s, &arg).into();
                 match arg.as_ref() {
-                    ";" => ContentString::HtmlUnsafe(cut.into()),
-                    _ => ContentString::HtmlSafe(cut.into()),
+                    ";" => ContentString::HtmlUnsafe(cut),
+                    _ => ContentString::HtmlSafe(cut),
                 }
             }
             ContentString::HtmlUnsafe(s) => ContentString::HtmlUnsafe(cut(s, &arg).into()),
--- a/src/render/filters.rs
+++ b/src/render/filters.rs
@@ -11,8 +11,9 @@ use crate::error::{AnnotatePyErr, PyRenderError, RenderError};
 use crate::filters::{
     AddFilter, AddSlashesFilter, CapfirstFilter, CenterFilter, CutFilter, DateFilter,
     DefaultFilter, DefaultIfNoneFilter, DivisibleByFilter, EscapeFilter, EscapejsFilter,
-    ExternalFilter, FilterType, LastFilter, LengthFilter, LowerFilter, SafeFilter, SlugifyFilter,
-    TitleFilter, UpperFilter, WordcountFilter, WordwrapFilter, YesnoFilter,
+    ExternalFilter, FilterType, ForceEscapeFilter, LastFilter, LengthFilter, LowerFilter,
+    SafeFilter, SlugifyFilter, TitleFilter, UpperFilter, WordcountFilter, WordwrapFilter,
+    YesnoFilter,
 };
 use crate::parse::Filter;
 use crate::render::common::gettext;
@@ -47,6 +48,7 @@ impl Resolve for Filter {
             FilterType::Escape(filter) => filter.resolve(left, py, template, context),
             FilterType::Escapejs(filter) => filter.resolve(left, py, template, context),
             FilterType::External(filter) => filter.resolve(left, py, template, context),
+            FilterType::ForceEscape(filter) => filter.resolve(left, py, template, context),
             FilterType::Last(filter) => filter.resolve(left, py, template, context),
             FilterType::Lower(filter) => filter.resolve(left, py, template, context),
             FilterType::Length(filter) => filter.resolve(left, py, template, context),
@@ -484,6 +486,18 @@ impl ResolveFilter for ExternalFilter {
             None => filter.call1((variable,))?,
         };
         Ok(Some(Content::Py(value)))
+    }
+}
+
+impl ResolveFilter for ForceEscapeFilter {
+    fn resolve<'t, 'py>(
+        &self,
+        variable: Option<Content<'t, 'py>>,
+        py: Python<'py>,
+        template: TemplateString<'t>,
+        context: &mut Context,
+    ) -> ResolveResult<'t, 'py> {
+        EscapeFilter.resolve(variable, py, template, context)
     }
 }
 
--- a/src/render/types.rs
+++ b/src/render/types.rs
@@ -471,10 +471,10 @@ impl<'t, 'py> Content<'t, 'py> {
         let output = match self {
             Self::Py(content) => resolve_python(content, context)?,
             Self::String(content) => content,
-            Self::Float(content) => ContentString::String(content.to_string().into()),
-            Self::Int(content) => ContentString::String(content.to_string().into()),
-            Self::Bool(true) => ContentString::String(Cow::Borrowed("True")),
-            Self::Bool(false) => ContentString::String(Cow::Borrowed("False")),
+            Self::Float(content) => return Ok(content.to_string().into()),
+            Self::Int(content) => return Ok(content.to_string().into()),
+            Self::Bool(true) => return Ok(Cow::Borrowed("True")),
+            Self::Bool(false) => return Ok(Cow::Borrowed("False")),
         };
         Ok(render_value_in_context(output, context))
     }
--- a/src/render/types.rs
+++ b/src/render/types.rs
@@ -456,16 +456,27 @@ pub enum Content<'t, 'py> {
     Bool(bool),
 }
 
+fn render_value_in_context<'t>(value: ContentString<'t>, context: &Context) -> Cow<'t, str> {
+    match value {
+        ContentString::String(content) | ContentString::HtmlSafe(content) => content,
+        ContentString::HtmlUnsafe(content) => match context.autoescape {
+            true => Cow::Owned(encode_quoted_attribute(&content).to_string()),
+            false => Cow::Owned(content.into()),
+        },
+    }
+}
+
 impl<'t, 'py> Content<'t, 'py> {
     pub fn render(self, context: &Context) -> PyResult<Cow<'t, str>> {
-        Ok(match self {
-            Self::Py(content) => resolve_python(content, context)?.content(),
-            Self::String(content) => content.content(),
-            Self::Float(content) => content.to_string().into(),
-            Self::Int(content) => content.to_string().into(),
-            Self::Bool(true) => "True".into(),
-            Self::Bool(false) => "False".into(),
-        })
+        let output = match self {
+            Self::Py(content) => resolve_python(content, context)?,
+            Self::String(content) => content,
+            Self::Float(content) => ContentString::String(content.to_string().into()),
+            Self::Int(content) => ContentString::String(content.to_string().into()),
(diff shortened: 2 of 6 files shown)


tests/filters/test_escape.py
```diff
@@ -63,3 +63,17 @@ def test_escape_autoescape_off_lower(assert_render):
...
+
+
+def test_autoescape_on_with_escape_does_nothing(template_engine):
+    template_without_escape = "{% autoescape on %}{{ html }}{% endautoescape %}"
+    template_with_escape = "{% autoescape on %}{{ html|escape }}{% endautoescape %}"
+    html = "<p>Hello World!</p>"
+    context = {"html": html}
+    rendered_without_escape = template_engine.from_string(
+        template_without_escape
+    ).render(context, None)
```
LilyFirefly: Do we need to explicitly pass the `None`? If so, can it have a keyword argument?

LilyFirefly (commented):
This is great!

I have one question/nitpick about one of the tests you added.

It would also be good to understand why `test_chaining08` from the Django suite is a `FAIL` instead of an `OK`. I suspect it's a different error message, in which case we can update the `scripts/django_tests_use_rusty_templates.patch` file to compensate.

UnknownPlatypus:
> It would also be good to understand why test_chaining08 from the Django suite is a FAIL instead of an OK. I suspect it's a different error message, in which case we can update the scripts/django_tests_use_rusty_templates.patch file to compensate.

fyi @Mouarius , the detailed logs are [here](https://github.com/LilyFirefly/django-rusty-templates/actions/runs/24606134520/job/71952598229?pr=409#step:8:12937)

tests/filters/test_escape.py
```diff
@@ -63,3 +63,17 @@ def test_escape_autoescape_off_lower(assert_render):
     html = "<p>Hello World!</p>"
     escaped = "&lt;p&gt;hello world!&lt;/p&gt;"
     assert_render(template=template, context={"html": html}, expected=escaped)
+
+
+def test_autoescape_on_with_escape_does_nothing(template_engine):
+    template_without_escape = "{% autoescape on %}{{ html }}{% endautoescape %}"
+    template_with_escape = "{% autoescape on %}{{ html|escape }}{% endautoescape %}"
+    html = "<p>Hello World!</p>"
+    context = {"html": html}
```
p-r-a-v-i-n:
```suggestion
    context = {"html": "<p>Hello World!</p>"}
```

tests/filters/test_escape.py
Mouarius: Indeed it can be removed.

src/render/filters.rs
```diff
@@ -217,7 +219,10 @@ impl ResolveFilter for CutFilter {
             ContentString::HtmlSafe(s) => {
                 let cut = cut(s, &arg);
```
LilyFirefly: I think we can move the `.into()` call up here:

```suggestion
                let cut = cut(s, &arg).into();
```

src/render/filters.rs
```diff
@@ -217,7 +219,10 @@ impl ResolveFilter for CutFilter {
...
                 match arg.as_ref() {
-                    ";" => ContentString::HtmlUnsafe(cut.into()),
+                    ";" => match context.autoescape {
+                        true => ContentString::HtmlUnsafe(cut.into()),
+                        false => ContentString::String(cut.into()),
+                    },
                     _ => ContentString::HtmlSafe(cut.into()),
```
```suggestion
                        true => ContentString::HtmlUnsafe(cut),
                        false => ContentString::String(cut),
                    },
                    _ => ContentString::HtmlSafe(cut),
```

tests/filters/test_escape.py
LilyFirefly:
```suggestion
    ).render(context)
```

tests/filters/test_escape.py
```diff
@@ -63,3 +63,17 @@ def test_escape_autoescape_off_lower(assert_render):
...
+    rendered_without_escape = template_engine.from_string(
+        template_without_escape
+    ).render(context, None)
+    rendered_with_escape = template_engine.from_string(template_with_escape).render(
+        context, None
```
LilyFirefly:
```suggestion
        context
```

LilyFirefly (commented):
This looks great! I just have a few code cleanup suggestions. Once they're resolved, I'm happy to approve/merge.

Mouarius:
> This is great!
> 
> I have one question/nitpick about one of the tests you added.
> 
> It would also be good to understand why `test_chaining08` from the Django suite is a `FAIL` instead of an `OK`. I suspect it's a different error message, in which case we can update the `scripts/django_tests_use_rusty_templates.patch` file to compensate.

I have found a quickfix for this issue, it patches the failing test but this is just a workaround to fix a deeper behaviour of the rusty template engine that defers from Django.
What I dislike about this patch, is that every filter implementation would need to have re-implement this logic for autoescaping.

# Investigation (bear with me, this is not AI generated)
To explain this, we have two interesting tests in Django's test suite: `test_chaining07` and `test_chaining08`.

## `test_chaining07`
### Input
- context: `a = "a < b"`
- template: `{{ a|force_escape|cut:";" }}`

**Notes:**
- `cut` removes all occurrences of the given character, so here `";" -> ""`
- by default, we are in a `autoescape` context, so `a` is expected to be escaped


### Current result (OK)
```
a &amp;lt b
```

### Explanation
Here is the Django pipeline:
| Step | Value | SafeString | Remarks |
| ------------- | -------------- |-------------- | -------------- |
| 0 | `a < b` | no |Initial, calls `VariableNode.render()`. |
| 1 - Filter resolution step | --- | --- |   First expression is `output = self.filter_expression.resolve(context)` at line `1077`. |
| 1.1 - apply `force_escape` | `a &lt; b` | yes | Applies `force_escape` filter |
| 1.2 - apply `cut:";"`| `a &lt b` | no | Cut filter changes a safe result to **unsafe** |
| 2 - Rendering node | --- | --- | After filter resolution, returns `render_value_in_context(output, context)`|
| 2 | `a &amp;lt b` | yes | Runs `conditional_escape()`, transforming the `&` in `&amp;` because `context.autoescape == True` and the string is **unsafe** |


Here is the `django-rusty-templates` pipeline
| Step | Value | Type | Remarks |
| ------------- | -------------- |-------------- | -------------- |
| 0 - variable resolution| `a < b` | `Content::Py` | `TokenTree::render` → `Render` impl for `Filter` → `Filter::resolve` recurses into `left`. Then `Variable::resolve` reads `a` from the context as a `Content::Python` obj. |
| 1 - apply `force_escape` | `a &lt; b` | `ContentString::HtmlSafe` | `ForceEscapeFilter::resolve` (`src/render/filters.rs`) calls `content.resolve_string(context)` → `resolve_python` makes `PyString` then `HtmlUnsafe("a < b")` -> `encode_quoted_attribute_to_string()` -> `ContentString::HtmlSafe`. |
| 2 - apply `cut:";"` | `a &lt b` | `ContentString::HtmlUnsafe` | `CutFilter::resolve` matches on `HtmlSafe` with arg `";"` and downgrades the output to `HtmlUnsafe` (`src/render/filters.rs:222`). This mirrors Django's `cut`, where removing `;` of a **safe** string makes it **unsafe** |
| 3 - render node | `a &amp;lt b` | --- | `Content::render` (`src/render/types.rs:462`) calls `ContentString::content()` which gets `ContentString::HtmlUnsafe` -> `encode_quoted_attribute` -> `&` becomes `&amp;`. |

## `test_chaining08`
### Input
- context: `a = "a < b"`
- template: `{% autoescape off %}{{ a|force_escape|cut:";" }}{% endautoescape %}`

**Notes:**
- same chain as `test_chaining07`, but wrapped in `{% autoescape off %}`

### Current result (FAIL)
```diff
- a &lt b
+ a &amp;lt b
?    ++++
```


### Explanation
Here is the Django pipeline:
| Step | Value | SafeString | Remarks |
| ------------- | -------------- |-------------- | -------------- |
| 0 | `a < b` | no |Initial, calls `VariableNode.render()`. |
| 1 - Filter resolution step | --- | --- |   First expression is `output = self.filter_expression.resolve(context)` at line `1077`. |
| 1.1 - apply `force_escape` | `a &lt; b` | yes | Applies `force_escape` filter |
| 1.2 - apply `cut:";"`| `a &lt b` | no | Cut filter changes a safe result to **unsafe** |
| 2 - Rendering node | --- | --- | After filter resolution, returns `render_value_in_context(output, context)`|
| 2 | `a &lt b` | no | `conditional_escape()` is a no-op because `context.autoescape == False` |

Here is the `django-rusty-templates` pipeline *before* the fix (what `test_chaining08` was failing on):
| Step | Value | Variant | Remarks |
| ------------- | -------------- |-------------- | -------------- |
| 0 | `a < b` | `Content::Py` | Same resolve path as above. |
| 1 — apply `force_escape` | `a &lt; b` | `ContentString::HtmlSafe` | Same as in `test_chaining07`; `force_escape` does not care about the autoescape context. |
| 2 — apply `cut:";"` | `a &lt b` | `ContentString::HtmlUnsafe` | Same branch at `src/render/filters.rs:222`. |
| 3 — render node (buggy) | `a &amp;lt b` | — | `ContentString::content()` implementation doesn't take `context` and `autoescape` into account, so it still calls `encode_quoted_attribute` which converts the `&lt` into `&amp;lt` again, causing a **FAIL** |


## Potential solution
We want somehow the `context` to be taken into account, like in the Django rendering mechanism, where the rendering pipeline is clear: 
1. Retrieve variable content
2. Resolve filters with `output = self.filter_expression.resolve(context)`
3. Render the result in context with `render_value_in_context(output, context)`

Current (if I understood well):
1. Retrieve variable content
2. Apply filters that return a certain type of `Content`
3. We render the result by calling the `render()` impl for `Content` that runs a different `resolve` impl for each possible `Content` type. 

### Suggestions
1. Change `content` impl of `Content` to take the `context: &Context` variable: straightforward, but maybe repetitive.
2. Create `render_value_in_context` method, and call it after resolving the different content types in `render` impl of `Content`

This commit implements the changes for option 2: https://github.com/LilyFirefly/django-rusty-templates/commit/08db5c3ffd6960f6d1ed3bb9aecb6c3b278299f4

### Questions
- Why did you split `ContentString` into `String | HtmlSafe | HtmlUnsafe` and not just into `Safe | Unsafe` ? I feel like `String` is a little bit misleading and adds strange branches.
- What do you think about the suggestions ?

LilyFirefly:
@Mouarius sorry I haven't got back to you about this - life has been pretty hectic the last week. It's definitely on my todo list to take a proper look.

Mouarius:
> @Mouarius sorry I haven't got back to you about this - life has been pretty hectic the last week. It's definitely on my todo list to take a proper look.

No worries, sorry @LilyFirefly I just wondered if you have seen my message, didn't want to urge nor bother you! Sorry for the aggressive review request!

LilyFirefly:
Hey @Mouarius! Sorry it took me so long to get to this! I've just read through your analysis and it's very clear! ~Can you add your `render_value_in_context` commit to this PR?~ Edit: Oh you already did this!

> Change `content` impl of `Content` to take the `context: &Context` variable: straightforward, but maybe repetitive.

Did you mean `ContentString`?

> Why did you split `ContentString` into `String | HtmlSafe | HtmlUnsafe` and not just into `Safe | Unsafe` ? I feel like String is a little bit misleading and adds strange branches.

I don't really remember why - maybe I wanted to avoid having to think about html all the time, since templates can be for any text format, but I think in hindsight it would be worthwhile seeing what a refactor to just `Safe` and `Unsafe` would look like.

src/render/types.rs
```diff
@@ -456,16 +456,27 @@ pub enum Content<'t, 'py> {
...
-            Self::Bool(true) => "True".into(),
-            Self::Bool(false) => "False".into(),
-        })
+        let output = match self {
+            Self::Py(content) => resolve_python(content, context)?,
+            Self::String(content) => content,
+            Self::Float(content) => ContentString::String(content.to_string().into()),
+            Self::Int(content) => ContentString::String(content.to_string().into()),
+            Self::Bool(true) => ContentString::String(Cow::Borrowed("True")),
+            Self::Bool(false) => ContentString::String(Cow::Borrowed("False")),
```
LilyFirefly: The compiler might do this for us, so I'm not sure there's a performance benefit, but I think it might still be more readable:

```suggestion
            Self::Float(content) => return Ok(content.to_string().into()),
            Self::Int(content) => return Ok(content.to_string().into()),
            Self::Bool(true) => return Ok(Cow::Borrowed("True")),
            Self::Bool(false) => return Ok(Cow::Borrowed("False")),
```

src/render/filters.rs
```suggestion
";" => ContentString::HtmlUnsafe(cut),
_ => ContentString::HtmlSafe(cut),
```

LilyFirefly (approved):
This looks great! I'm going to apply my suggested changes and merge! Thanks again, and I'm sorry it took so long to give this a proper review!

PR #410, opened 2026-04-18
Get Django's template_loader tests passing

LilyFirefly:
Add a stub `Origin` class into the `TemplateDoesNotExist.tried` list. Avoid expecting the `TemplateDoesNotExist.backend` attribute to be set, since this is optional.

### DIFF
--- a/scripts/django_tests_use_rusty_templates.patch
+++ b/scripts/django_tests_use_rusty_templates.patch
@@ -27,6 +27,87 @@ index f1a194d..0b27dd4 100755
      from django.apps import apps
      from django.conf import settings
      from django.core.exceptions import ImproperlyConfigured
+diff --git a/tests/template_loader/tests.py b/tests/template_loader/tests.py
+index 569f86b..c8b488c 100644
+--- a/tests/template_loader/tests.py
++++ b/tests/template_loader/tests.py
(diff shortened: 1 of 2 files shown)


scripts/django_tests_use_rusty_templates.patch
```diff
@@ -27,6 +27,113 @@ index f1a194d..0b27dd4 100755
...
+         self.assertEqual(e.exception.chain[2].backend.name, "dummy")
+-        self.assertEqual(
+-            e.exception.chain[3].tried[0][0].template_name,
+-            "template_loader/missing.html",
++        self.assertTrue(
++            e.exception.chain[3].tried[0][0].template_name.endswith(
++                "template_loader/missing.html",
++            )
+         )
+-        self.assertEqual(e.exception.chain[3].backend.name, "django")
```
UnknownPlatypus, line 107: Why do we need this change ? How is rusty temlpate displaying the error ?

Also the ecossytem check comment seem to have an issue displaying the passing rate 🤔

LilyFirefly: We don't set the `backend` attribute, so this was an `AttributeError: 'NoneType' object has no attribute 'name'`.

PR #411, opened 2026-04-18
Improve `justfile`

Mouarius:
Added some *nice to have* recipes and features to the justfile to facilitate onboarding and compatibility.

- [x] `bootstrap` recipe: includes the `compilemessages` command, as referenced in the `CONTRIBUTING.md` guide.
- [x] list behaviour on `just` empty call
- [x] `python-test` to run `maturin` then `pytest` 
- [x] `set -euox pipefail` as recommended in the [official docs](official docs)
- [x] `/usr/bin/env bash` instead of `/usr/bin/bash` in scripts

These are just suggestions, please add some others if you find them useful!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,9 +1,12 @@
 ## Getting started with development
 
-django-rusty-templates is written in Rust, so you'll need to install the
-[Rust toolchain](https://www.rust-lang.org/tools/install).
+### Requirements
+`django-rusty-templates` is written in Rust, so you'll need to install the [Rust toolchain](https://www.rust-lang.org/tools/install).
 
-### Option 1: Using uv
+To run the `just` recipes, you will need to have [Just](https://github.com/casey/just) installed.
+
+### Python dependencies
+#### Option 1: Using uv
 
 [uv](https://docs.astral.sh/uv/) is a fast Python package installer and resolver. Using it will significantly speed up dependency installation.
 
@@ -20,7 +23,7 @@ $ uv sync
 $ source .venv/bin/activate
 ```
 
-### Option 2: Using standard Python tools
+#### Option 2: Using standard Python tools
 
 If you prefer not to use uv, you can set up your development environment with standard Python tools:
 
@@ -32,6 +35,9 @@ $ pip install --group dev
 
 Note: The `[dev]` dependency group is defined in `pyproject.toml` and includes all necessary development dependencies.
 
+### Final step
+Run `just bootstrap`.
+
 ## Running tests
 
 ### Python tests with pytest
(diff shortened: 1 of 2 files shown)


justfile
```diff
@@ -1,18 +1,45 @@
...
+# Bootstrap the environment for running the test suite
+bootstrap:
+    #!/usr/bin/env bash
+    set -euox pipefail
+    # Compile translations
+    cd tests && django-admin compilemessages
+
+# Build the Rust extension and run the Python test suite (extra args forwarded to pytest)
+[group('test')]
+python-test *ARGS:
```
UnknownPlatypus, line 16: This is great, I often forget maturin otherwise

justfile
```diff
@@ -1,18 +1,45 @@
+set shell := ["bash", "-euox", "pipefail", "-c"]
+
+# List available recipes
+_default:
+    @just --list --unsorted
+
+# Bootstrap the environment for running the test suite
+bootstrap:
+    #!/usr/bin/env bash
+    set -euox pipefail
```
LilyFirefly, line 10: Do we need these as well as setting `shell` at the top of the file? It feels repetitive, but maybe it's necessary?

CONTRIBUTING.md
```diff
@@ -32,6 +35,9 @@ $ pip install --group dev
 
 Note: The `[dev]` dependency group is defined in `pyproject.toml` and includes all necessary development dependencies.
 
+### Final step
+Run `just bootstrap`.
```
LilyFirefly: I think we should spell out the manual `compilemessages` step here too, for people who don't use `just`.

CONTRIBUTING.md
```diff
@@ -42,7 +48,7 @@ $ maturin develop
 $ pytest
 ```
 
-If translation tests are failing, make sure that you have compiled django translations by running `django-admin compilemessages` in `tests/` directory.
+Or you can run `just python-test`
```
LilyFirefly:
```suggestion
Running `just python-test` combines these two commands for you.
```

LilyFirefly (commented):
These are all great ideas! I have a few questions and suggestions.

justfile
Mouarius: Yes unfortunately, the shell directive only applies to command recipes, not scripts (starting with shebang)
This is described in the [docs](https://just.systems/man/en/configuring-the-shell.html?highlight=shell#configuring-the-shell)

CONTRIBUTING.md
Mouarius: You're right, let me update it!

LilyFirefly (approved):
Amazing!

PR #492, opened 2026-08-22
Implemented cycle tag.

p-r-a-v-i-n:
refs #2


LilyFirefly:
Thanks for this! Can you take a look at the failing upstream tests (https://github.com/LilyFirefly/django-rusty-templates/pull/492#issuecomment-5381418886) and work out if these are genuine issues with this implementation or just something we need to patch the tests to fix?

p-r-a-v-i-n:
> Thanks for this! Can you take a look at the failing upstream tests ([#492 (comment)](https://github.com/LilyFirefly/django-rusty-templates/pull/492#issuecomment-5381418886)) and work out if these are genuine issues with this implementation or just something we need to patch the tests to fix?

I will check.

p-r-a-v-i-n:
1. I took some tests from django for cycle tag.
django uses `filter`, `ifchanged`, `with`, `resetcyle`  in the tests. so currently i was expecting those tests to fail.
Should i remove them for now ? may be we should

LilyFirefly:
Ah, if those tests rely on other tags we haven't implemented, that's OK for now. Can you link to the Django source code for those tests?

p-r-a-v-i-n:
also one more thing , translation error

   ```
@setup(
        {
            "i18n14": (
                '{% cycle "foo" _("Password") _(\'Password\') as c %} {% cycle c %} '
                "{% cycle c %}"
            )
        }
    )
    def test_i18n14(self):
        with translation.override("de"):
            output = self.engine.render_to_string("i18n14")
        self.assertEqual(output, "foo Passwort Passwort")

```

currently our rusty produces `foo Password Password`, so i think this should be fixed in seperate PR.

**Comment by LilyFirefly** [comment]
Yeah, translation support is also currently incomplete.

File: tests/tags/test_cycle.py [hunk 1]
```diff
@@ -1,6 +1,8 @@
-import pytest
 from inline_snapshot import snapshot
 
+# Some tests taken from Django tests suite for cycle tag:
+# https://github.com/django/django/blob/main/tests/template_tests/syntax_tests/test_cycle.py
```
LilyFirefly: Given we're currently targeting Django 5.2, we should link to 5.2:

```suggestion
# https://github.com/django/django/blob/stable/5.2.x/tests/template_tests/syntax_tests/test_cycle.py
```

LilyFirefly: Though I prefer to write our own tests rather than copying Django's - we have https://github.com/LilyFirefly/django-rusty-templates/pull/492#issuecomment-5381418886 to regression-test against Django, so copying tests is not really necessary.

p-r-a-v-i-n: For now or for this PR lets keep them. Additionally we can add more tests. 
Initially i had only 3 or 4 tests so i quickly i wanted more scenarios . so i always take reference or just mimic those django tests

tests/tags/test_cycle.py
```diff
@@ -0,0 +1,409 @@
...
+        expected=expected,
+    )
+
+
+def test_cycle_missing_argument_error(assert_parse_error):
+    assert_parse_error(
+        template="{% cycle %}",
+        django_message="'cycle' tag requires at least two arguments",
+        rusty_message=snapshot("""\
+  × Expected an argument
```
LilyFirefly:
```suggestion
  × Expected two or more arguments
```

tests/tags/test_cycle.py
```diff
@@ -0,0 +1,409 @@
...
+    class Unstringable:
+        def __str__(self):
+            raise ValueError("cycle rendering error")
+
+    assert_render_error(
+        template="{% cycle value 'fallback' %}",
+        context={"value": Unstringable()},
+        exception=ValueError,
+        django_message=snapshot("cycle rendering error"),
+        rusty_message=snapshot("cycle rendering error"),
```
LilyFirefly, line 233: Can we annotate this with `miette` for better debugging? Or is there a reason that's difficult?

tests/tags/test_cycle.py
```diff
@@ -0,0 +1,409 @@
...
+            "items": [1, 2, 3, 4],
+            "included_template": included_template,
+        },
+        expected="abca",
+    )
+
+
+def test_cycle_single_value_as_syntax(assert_render):
+    assert_render(
+        template="{% cycle value as current %}",
```
LilyFirefly, line 461: This is a bit odd! I'd expect to see the "'cycle' tag requires at least two arguments" error we get without the `as current` part. Perhaps this is a minor upstream Django bug?

tests/tags/test_cycle.py
LilyFirefly, line 1: Can we have a test like this:

```python
def test_nested_named_cycles(assert_render):
    assert_render(
        template='{% cycle "a" "b" as foo %}{% cycle foo "c" as bar %}{% cycle bar %}{% cycle bar %}{% cycle foo %}{% cycle bar %}{% cycle bar %}',
        context={},
        expected="aacabcb",
    )
```

tests/tags/test_cycle.py
LilyFirefly: Can we also have a test for `{% cycle "a" %}`? I think we'd like to handle this as a not-enough arguments error, rather than an unknown cycle error.

dtl-lexer/src/tag/cycle.rs
```diff
@@ -0,0 +1,100 @@
...
+pub enum CycleArguments {
+    Reference {
+        name: At,
+    },
+    Definition {
+        values: Vec<TagElementToken>,
+        name: Option<At>,
+        silent: bool,
+    },
+}
```
LilyFirefly: Distinguishing between a `Reference` and a `Definition` is a parsing-level concern. The lexer should be focused on creating a sequence of tokens.

A tag like:
```html
{% cycle "a" b 1.2 as foo silent %}
```
would become a sequence of tokens such as:
* `Text`
* `Variable`
* `Numeric`
* `Variable`
* `Variable`
* `Variable`

Given this is the structure that `TagElementLexer` already gives us, I think we don't need a custom `CycleLexer` and can move all the logic in this file into the parser. This might also allow some scope for simplification in the parser.

src/parse.rs
```diff
@@ -720,6 +726,17 @@ pub struct FirstOf {
     pub asvar: Option<String>,
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct Cycle {
+    pub id: CycleId,
+    pub values: Vec<TagElement>,
```
LilyFirefly: We never need to mutate `values` after creating `Cycle`, so we can optimise this by using `Box<[TagElement]>`:

```suggestion
    pub values: Box<[TagElement]>,
```

This stores the `values` as a fixed-size array on the heap, rather than a dynamically-sized array. This saves the memory that `Vec` uses to store its `capacity` and the memory of any unused heap space the `Vec` is keeping available for new elements.

The method [`into_boxed_slice`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.into_boxed_slice) is useful for turning a `Vec<T>` into a `Box<[T]>`.

src/render/tags.rs
```diff
@@ -1390,3 +1391,34 @@ impl Render for FirstOf {
...
+        template: TemplateString<'t>,
+        context: &mut Context,
+    ) -> RenderResult<'t> {
+        let index = context.next_cycle_index(self.id, self.values.len());
+        let value = &self.values[index];
+        let content = value.resolve(py, template, context, ResolveFailures::Raise)?;
+
+        let Some(content) = content else {
+            if let Some(asvar) = &self.asvar {
+                context.insert(asvar.clone(), PyString::new(py, "").into_any());
```
LilyFirefly: We can avoid creating a new empty string on the Python heap every time with [`intern!`](https://pyo3.rs/main/doc/pyo3/macro.intern):

```suggestion
                context.insert(asvar.clone(), intern!(py, "").into_any());
```

src/render/tags.rs
```diff
@@ -1390,3 +1391,34 @@ impl Render for FirstOf {
...
+            return Ok(Cow::Borrowed(""));
+        };
+
+        if let Some(asvar) = &self.asvar {
+            context.insert(asvar.clone(), content.to_py(py));
+        }
+
+        if self.silent {
+            return Ok(Cow::Borrowed(""));
+        }
```
LilyFirefly: We have a bunch of branching based on whether `asvar` and/or `silent` are set. It might be a nice refactor to split `Cycle` into three types (`SimpleCycle`, `NamedCycle` and `SilentNamedCycle`) at the parsing stage. Then we could avoid the runtime checks for behaviour here.

src/render/types.rs
```diff
@@ -294,6 +298,13 @@ impl Context {
             }
         }
     }
+
+    pub fn next_cycle_index(&mut self, id: CycleId, length: usize) -> usize {
+        let index = self.cycle_indices.entry(id).or_insert(0);
+        let current = *index;
+        *index = (current + 1) % length;
+        current
```
LilyFirefly: I think this is a little more idiomatic:

```suggestion
        *self
            .cycle_indices
            .entry(id)
            .and_modify(|index| *index = (*index + 1) % length)
            .or_default()
```

LilyFirefly (commented):
Thanks again for the PR, great work! I have some suggestions below, but this is a great starting point.

dtl-lexer/src/tag/cycle.rs
p-r-a-v-i-n: This is something i was expecting , so thanks. I wasn't sure to for its location.

src/render/tags.rs
p-r-a-v-i-n: This is new learning.

src/parse.rs
```diff
@@ -862,6 +880,13 @@ pub enum ParseError {
         #[label("here")]
         at: SourceSpan,
     },
+    #[error("Unknown named cycle '{name}'")]
```
LilyFirefly: I think some more Django tests will pass if we match this closer to Django's message:

```suggestion
    #[error("Named cycle '{name}' does not exist")]
```

We'll need to adjust our tests too of course, but I think Django's message is pretty good here so there's no reason to diverge.

tests/tags/test_cycle.py
p-r-a-v-i-n: should i remove this test ? django behaviour is even more strange. 
for example,
`{% cycle value as current %}` here django does not recognize the `as current` as assignment.

p-r-a-v-i-n: should we raise this upstream ?

LilyFirefly: > should we raise this upstream ?

Yeah, I think that would be a good idea.

p-r-a-v-i-n: I didn't understand fully but does this [ticket](https://code.djangoproject.com/ticket/5865) talking about that problem ? when you get some spare time, do you mind going through that ticket ?

p-r-a-v-i-n: i wrote this [comment](https://code.djangoproject.com/ticket/5865#comment:28) in that ticket.

tests/tags/test_cycle.py
```diff
@@ -234,6 +234,60 @@ def __str__(self):
...
+    class Unstringable:
+        def __str__(self):
+            raise ValueError("cycle rendering error")
+
+    assert_render_error(
+        template="{% cycle value 'fallback' as current %}",
+        context={"value": Unstringable()},
+        exception=ValueError,
+        django_message=snapshot("cycle rendering error"),
+        rusty_message=snapshot("cycle rendering error"),
```
LilyFirefly: Is there a reason we can't have source annotation for this error?

src/parse.rs
```diff
@@ -1534,6 +1588,80 @@ impl<'t, 'py> Parser<'t, 'py> {
...
+                    ParseError::UnknownNamedCycle {
+                        name: name_text.to_string(),
+                        at: reference.at.into(),
+                    }
+                })
+            }
+            [argument] => Err(ParseError::MissingCycleArguments {
+                at: argument.at.into(),
+            }),
+            tokens => {
```
LilyFirefly: If we return early in the first three match arms, we can dedent the big bit of logic in the fourth arm:

```suggestion
        let tokens = match tokens.as_slice() {
            [] => return Err(ParseError::MissingCycleArguments { at: at.into() }),
            [reference] if reference.token_type == TagElementTokenType::Variable => {
                let name_text = self.template.content(reference.at);

                return self.named_cycles.get(name_text).cloned().ok_or_else(|| {
                    ParseError::UnknownNamedCycle {
                        name: name_text.to_string(),
                        at: reference.at.into(),
                    }
                })
            }
            [argument] => return Err(ParseError::MissingCycleArguments {
                at: argument.at.into(),
            }),
            tokens => tokens,
```

src/parse.rs
```diff
@@ -1534,6 +1588,80 @@ impl<'t, 'py> Parser<'t, 'py> {
...
+                let (tokens, name, silent) = match tokens {
```
LilyFirefly: I think `name_at` would be a slightly more readable name:

```suggestion
                let (tokens, name_at, silent) = match tokens {
```

src/parse.rs
```diff
@@ -1534,6 +1588,80 @@ impl<'t, 'py> Parser<'t, 'py> {
...
+                    }
+                    values => (values, None, false),
+                };
+
+                let mut values = Vec::with_capacity(tokens.len());
+
+                for token in tokens {
+                    let value = token.parse(self)?;
+                    values.push(value);
+                }
```
LilyFirefly, line 1635:
```suggestion
                        let values = tokens
                            .iter()
                            .map(|token| token.parse(self))
                            .collect::<Result<_, _>>()?;
```

LilyFirefly (approved):
This is great! Just a few minor tweaks and this can be merged!

tests/tags/test_cycle.py
p-r-a-v-i-n: Please correct me. I assume you mean here can we do this without writing a special error class? Actually i didn't thought in that way. I will try

LilyFirefly: I mean, can we have the source code annotation, like we do for the other errors. I'm hoping it would be possible to copy the approach used elsewhere, but maybe there's a reason it's difficult here?

p-r-a-v-i-n: Ah. Sorry I got you yeah i think that's possible.

PR #498, opened 2026-09-03
Bump rust 1.98

p-r-a-v-i-n:
may be followup from https://github.com/LilyFirefly/django-rusty-templates/pull/490


UnknownPlatypus:
It should be possible to teach renovate about it with a group so it bumps both in one pr ?

LilyFirefly (approved):

PR #501, opened 2026-09-05
Allow choosing which Django tests to run at the command line

LilyFirefly:
This is useful when adding a new feature to avoid needing to scroll through the full test suite output to find the tests of interest.


UnknownPlatypus (approved):
Looks convenient 👍

